PR #3, opened 2017-07-24
Todo List Jake/Antonio

Chooster:


src/index.js
```diff
@@ -1,7 +1,48 @@
...
+    });
+  }
+  onClear() {
+    this.setState({
+      list: []
+    });
+  }
+  onCompleted(idx) {
+    let newList = this.state.list;
+    newList[idx].completed = !newList[idx].completed;
```
ryanhca: Nicely done! This is a great implementation of toggling a boolean state on an object. You'll see these toggle functions everywhere in React development

src/TodoInputField.js
```diff
@@ -0,0 +1,39 @@
...
+      text: '',
+    };
+    this.handleSubmit = this.handleSubmit.bind(this);
+    this.handleChange = this.handleChange.bind(this);
+  }
+  handleChange(e) {
+    this.setState({ text: e.target.value });
+  };
+
+  handleSubmit(e) {
```
ryanhca: This is an interesting approach. Normally I would wait for the user to be done typing and have them physically hit "Enter" or click "Submit" for this to be fired.

PR #4, opened 2017-07-24
cleaned it up to look nearly identical to Ben's, still not great

dwinslow123:


src/components/App.js
```diff
@@ -0,0 +1,14 @@
+import React, { Component } from 'react';
+import TodoForm from './TodoForm';
+
+class App extends Component {
+  render() {
+    return (
+        <div>
```
ryanhca: I'm not positive you need this `<div>` here. Because react render method can return a single element.

src/components/App.js
```diff
@@ -0,0 +1,14 @@
+import React, { Component } from 'react';
+import TodoForm from './TodoForm';
```
ryanhca: If you don't need state, you don't need your component to be a class, you can turn it into a function... if ya want to.

src/components/TodoList.js
```diff
@@ -0,0 +1,13 @@
+import React, { Component } from 'react';
+
+export default (props) => {
```
ryanhca: You should name the components you create instead of exporting out anonymous functions. If something breaks down in this, you'll have a hell of a time running around stack trace errors.

PR #9, opened 2017-08-01
Completed assignment

Egrodo:
I'm late because I wasn't able to make class last weekend, but here I'm catching up. Here I've completed the to-do list assignment (and went overboard with styling and extras).

### DIFF
--- a/src/List.js
+++ b/src/List.js
@@ -1,0 +1,32 @@
+import React from 'react';
+
+export default ({ ToDos }) => { // Export an annonymous function that takes in the ToDos we have and displays them in a list.
+
+    const clickHandler = (item) => {
+      item.completed = !item.completed;
+      const checkbox = document.getElementById(`${item.text}checkbox`);
+      const text = document.getElementById(`${item.text}`);
+      if (item.completed) {
+        checkbox.checked = true;
+        text.classList.add('crossed');
+        console.log(`'${item.text}' marked done.`);
+      } else {
+        checkbox.checked = false;
+        text.classList.remove('crossed');
+        console.log(`'${item.text}' marked not done.`)
+      }
+    };
+
+    return ( // Any time you want to use JavaScript inside JSX wrap in {}.
+      <ul>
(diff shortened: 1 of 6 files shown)


src/List.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+export default ({ ToDos }) => { // Export an annonymous function that takes in the ToDos we have and displays them in a list.
+
+    const clickHandler = (item) => {
+      item.completed = !item.completed;
+      const checkbox = document.getElementById(`${item.text}checkbox`);
```
ryanhca: THere are actually much easier ways to do this than to selected the DOM. I would research how checkBoxes are handled in react most commonly.

src/List.js
```diff
@@ -0,0 +1,32 @@
...
+      } else {
+        checkbox.checked = false;
+        text.classList.remove('crossed');
+        console.log(`'${item.text}' marked not done.`)
+      }
+    };
+
+    return ( // Any time you want to use JavaScript inside JSX wrap in {}.
+      <ul>
+        {ToDos.map((ToDo, i) => { // For each value inside ToDos TODO why map instead of forEach?
```
ryanhca: .map returns a copy of the original array where as forEach is better for looping over an array and mutating the original copy. It's a common pattern in react to use .map for all the things UI.

src/ToDo.js
```diff
@@ -0,0 +1,68 @@
...
+
+  addToDo(event) {
+    event.preventDefault(); // Disable the default behavior of refreshing the page upon clicking enter.
+    document.getElementById("mainInput").value = ""; // Clear input field.
+
+    if (!previousToDos.includes(this.state.newToDo)) { // Check for duplicates.
+      console.log(`Adding '${this.state.newToDo}' to the list.`);
+      previousToDos.push(this.state.newToDo);
+
+      if (document.getElementById('response').innerHTML) document.getElementById('response').innerHTML = ''; // If the response box needs to be cleared, do so.
```
ryanhca: once again there are easier patterns for selecting elements in React vs accessing the DOM directly.

PR #13, opened 2017-08-25
Completed LS-React-1 + EC WIP

lefrenk:
Did 1 of 3 of the extra credit.


src/components/toDoInput.js
```diff
@@ -0,0 +1,49 @@
...
+  borderBottom: '1px solid #ddd',
+  backgroundColor: ' #f5f5f5',
+  marginLeft: 10
+};
+
+// Class Component
+class ToDoInput extends React.Component {
+  constructor(props) {
+    super(props);
+    // this.state.value is <input> value
```
ryanhca: Comments  should be removed before submitting PRs

src/components/toDoInput.js
```diff
@@ -0,0 +1,49 @@
...
+    this.state = { value: '' };
+    this.handleChange = this.handleChange.bind(this);
+    this.handleSubmit = this.handleSubmit.bind(this);
+  }
+  handleChange(event) {
+    this.setState({ value: event.target.value });
+  }
+  handleSubmit(event) {
```
ryanhca: Instead of binding this to a keyDown event, you might find it easier to bind it to an onSubmit event. That way you don't have to do any logic to guess the key.

PR #15, opened 2017-08-26
LS-React-I

username1001:


src/components/App.js
```diff
@@ -0,0 +1,49 @@
...
+class App extends Component {
+  constructor() {
+    super();
+    this.state = {
+      list: [] // Need to push items into this array.
+    };
+  }
+  addItem() {
+    const item = document.getElementById('listItem').value;
+    document.getElementById('listItem').value = '';
```
ryanhca: There are a lot of better ways to get things done in React than this. I would recommend looking at how to do things without having to select the 'document'

src/components/App.js
```diff
@@ -0,0 +1,49 @@
+import React, { Component } from 'react';
+import { render } from 'react-dom';
+
+const ListItem = (props) => {
```
ryanhca: Maybe spend some time and look at how to break this out into another file. That's sort of the react way of doing things.

PR #16, opened 2017-08-26
react to do list, no styling

khamp19:
no styling, no update or delete

### DIFF
--- a/src/to_do_components/addToDo.js
+++ b/src/to_do_components/addToDo.js
@@ -1,0 +1,18 @@
+import React, { Component } from 'react';
+// form doesnt reset after submit?
+
+const TodoForm = (props) => {
+    return(
+      <ul>
+        <input id="newiteminput"
+          type='text'
+          placeholder="add new todo"
+          value={props.newTodo}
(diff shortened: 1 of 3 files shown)


src/to_do_components/addToDo.js
```diff
@@ -0,0 +1,18 @@
+import React, { Component } from 'react';
+// form doesnt reset after submit?
```
ryanhca: one way you could make this work is to set the state of your string back to ""

src/to_do_components/todoList.js
```diff
@@ -0,0 +1,47 @@
...
+  this.makeNewTodo = this.makeNewTodo.bind(this)
+  this.addTodo = this.addTodo.bind(this);
+  };
+
+  makeNewTodo(event) {
+    this.setState({newTodo: event.target.value});
+  }
+
+  addTodo() {
+    const todos = this.state.todos;
```
ryanhca: This looks awesome. But you could simplify this to one line like `const { todos, todo, newTodo} = this.state`

PR #17, opened 2017-08-27
Todo app complete (for now)

sam-crabtree:
Plan on extending the functionality of this, but for now you can add/remove items.

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -1,0 +1,20 @@
+import React, { Component } from 'react';
+import Heading from './Heading';
+import TodoForm from './TodoForm';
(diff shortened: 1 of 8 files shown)


src/components/App.js
```diff
@@ -0,0 +1,20 @@
...
+export default class App extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {}
+  }
+
+  render() {
+    return (
+      <div className='div-container'>
+        <Heading />
```
ryanhca: I like this! Two components that are siblings. Very clean very nice.

PR #18, opened 2017-08-28
Made some minor mistakes

ericblancas23:
I focused more on completing the project through codesandbox just to help me with the visuals. Won't happen again


src/App.js
```diff
@@ -0,0 +1,57 @@
...
+    this.addTodo = this.addTodo.bind(this);
+    this.removeTodo = this.removeTodo.bind(this);
+  }
+
+  addTodo(todoText) {
+    let todos = this.state.todos.slice();
+    todos.push({id: this.state.nextId, text: todoText});
+    this.setState({
+      todos: todos,
+      nextId: ++this.state.nextId
```
ryanhca: I love this idea of managing the TodoId as a counter!

src/components/todoInput.js
```diff
@@ -0,0 +1,33 @@
...
+    this.handleChange = this.handleChange.bind(this);
+    this.addTodo = this.addTodo.bind(this);
+  }
+
+  handleChange(e) {
+    this.setState({value: e.target.value});
+  }
+
+  addTodo(todo) {
+    // Ensure a todo was actually entered before submitting
```
ryanhca: Try and remove all comments before submitting your Pull -requests

src/components/todoItem.js
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  removeTodo(id) {
+    this.props.removeTodo(id);
+  }
+
+  render() {
+    return (
+      <div className="todoWrapper">
+        <button className="removeTodo" onClick={(e)=> this.removeTodo(this.props.id) }>remove</button>{this.props.todo.text}
```
ryanhca: using a lambda is a good way to only bind the function you're calling to the element in the render method that you're using to call it with. However, a pattern that I like, (since you have a constructor) is to bind the handlers in the constructor . You also are passing in "e" and not using it. I'd advice against this in the future.

PR #20, opened 2017-09-25
Didn't get into Sass

coleferg:


src/app.js
```diff
@@ -0,0 +1,97 @@
...
+        const item = this.state.newItem
+        todos.push(item);
+        this.setState({
+            todos: todos,
+            newItem: {
+                text: '',
+                completed: false
+            }
+        })
+        localStorage.setItem("todos", JSON.stringify(todos));
```
ryanhca: I like that you are using localStorage to persist some data. But Component state is just fine for a todo list built with the client.

PR #40, opened 2017-10-23
Adding tasks works, started work on removing (commented it out)

perryahern:


src/components/App.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  handleAddTaskToList() {
+    let todo = this.state.toDoTask;
+    const toDos = this.state.toDos;
+    toDos.push(todo);
+    this.setState({ toDos, toDoTask: '' });
+  }
+
+  // handleRemoveTaskFromList() {
```
ryanhca: The only thing you need to do to fix this is to have some sort of id to compare the task you want deleted with the tasks in the array. Then run the splice over where the item is found at that index

PR #41, opened 2017-10-24
Cassidy Avery functioning w/out EC - Paired w/ Don Utley

tacolim:


src/components/toDo.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+const ToDo = props => {
```
ryanhca: I love this. Single reusable functional component! Beautiful!

PR #98, opened 2018-02-05
Finished all questions in Answers.md (Cliff Kang)

cliffkang:


mixelpixel:
@cliffkang nicely written. Reads like you have absorbed the material and have a good handle on the subject matter 👍

PR #120, opened 2018-02-06
Christopher Beards

ChristopherBeards:
I had a hard time with imports/exports and so created the Todo App on one page. I ended up having to start a new React Project, and then copy/paste the files over when I made some progress. Otherwise I would have had a steady stream of commits.


manishavi:
Hey Christopher,
Good to see your pull request and your name on it. You are good for the commits in this one. answers file shows that you have gone through the docs and have a good understanding of the concepts.I downloaded your project and ran it with all the dependencies, and it was only giving one console error as you were missing the key inline 50. you can solve it by adding the key like  `{this.state.list.map( (value,i) => <li key={i} className="li">{value}</li>)}` .
Nice design and good functionality.

Thanks

PR #124, opened 2018-02-06
Punit Rawal: react-I..work-in-progress

rawalpunit:
So...This is hard work..what you're doing..here's something to lighten the mood.

Bear walks into a bar and says "I'll have fish'n'chips................................................................and a beer"
Bartender: sure..but why the long pause
Bear: Oh these? I was born with 'em

Oh..forgot about this....Daniel Lara..and Lo Saephan and I worked together on this though I think we're submitting individually after writing up our own code.

### DIFF
--- a/src/components/Todo.js
+++ b/src/components/Todo.js
@@ -1,0 +1,17 @@
+import React from 'react';
+
+function Todo(props) {
+  return (
+    <div>
+      {props.todos.map((todo,i) => {
+        if(todo.complete) {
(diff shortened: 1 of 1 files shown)


satishvattikuti:
Lol, Thanks for the joke. Haha.

I was reviewing your code. I cloned it to my machine and ran it. I only see a header with ToDO  in it. There is no other component. did you forget to push it ?.

PR #128, opened 2018-02-07
Nathaniel Flory - To do List

Nate152:


satishvattikuti:
Very well written code. Neatly organized. I ran it in my system. A few console errors
1. You might have forgotten to use key property in Li item. like#  <li key = {}> whatevr </li> 
2. the class is called as className when it comes to react css.

May be that's why I do not see any stylying. fix that and add some stylying if you can. 

Make sure to keep all the reusable code pieces in one places,like components.

It's working nice. You can add the tasks to list. You did a wonderfull job. Keep up the good work.

PR #141, opened 2018-03-05
Working on the React-I

adfaris:
Working on the List Project, We were able to accept task and append it at the end.


AndjDavis:
Nice work guys, good job with the callback function in the deleteTodo.  Everything works when I pull it up in the browser.  I see you're still working on LocalStorage, apart from everything looks great!

PR #146, opened 2018-03-05
Justin Kaseman - Complete (Stretch WIP)

Jkasem:
- [x] answers.md
- [x] todo item
- [x] todo
- [ ] stretch:   
* Refactor each todo to be an object instead of a just a string. For example, `todo: {'text': 'Shop for food, 'completed': false}` and when a user clicks on a todo, switch that completed flag to true. If `completed === true`, this should toggle the strikethrough on the 'completed' todo. The toggling functionality should work the same as when each todo was just a string. 
  * Add the ability to delete a todo. The way this would work is each todo item should have an 'x' that should be clickable and that, when clicked, should remove the todo item from the state array, which will also remove it from the rendered list of todos. 
  * Take your App's styles to the next level. Start implementing as much creativity here as you'd like. You can build out your styles on a component-by-component basis eg `App.js` has a file next to it in the directory tree called `App.scss` and you define all your styles in that file. Be sure to @import these styles into the `index.scss` file. 
  * Persist your data in `window.localStorage()` hint: you may have to pass your data to a stringifier to get it to live inside the `localStorage()` of the browser. This will cause it to persist past the page refresh.

### DIFF
--- a/answers.md
+++ b/answers.md
@@ -1,19 +1,20 @@
 ### All Answers to Partner Study should be filled out in this file.
 
-1. Single Page Application
-
+1. Single Page Application (SPA)
+    Loads a single HTML file and dynamically update as the user interacts with them. Much of the work happens client side in the browser.
 2. Compilers
-
+    Software that transforms one computer language to another computer language.
 3. Bundlers
-
+    A tool that analyze's complex projects and compiles them into something the browser can understand.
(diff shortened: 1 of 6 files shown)


Jkasem:
http://colorful-snails.surge.sh/

PR #147, opened 2018-03-05
Randy Wilson, Nour Hassan

Phoenixrr2113:


mvpache:
Good work. If you get a chance, see if you can get toggling complete to control strike-through.

Also I'd like to see cleared commit messages, detailing what feature you worked on etc

PR #148, opened 2018-03-05
Jake Henry, AJ Genung, Walter Woodward React-I

Blast3d:


mvpache:
Killer work on the callback functions in the child. Keep it up. See if you can push yourself on component styling next time you are available for a stretch, there are a few ways to do this.

PR #159, opened 2018-03-06
WIP - Courtney Seitz

CourtneySeitz:
All completed except for the extra credit items.


mvpache:
Good work Courtney. Couple things though: classes in JSX are defined with "className", because "class" is reserved by JS. Also you'll want to name your components descriptive things rather than child/parent/app. "TodoList" and "TodoItem" would have been an example for this. Otherwise, great, keep it up.

PR #169, opened 2018-03-22
Caroline Cariste: Solution in progress

pinto8:


answers.md
```diff
@@ -1,19 +1,64 @@
 ### All Answers to Partner Study should be filled out in this file.
 
 1. Single Page Application
+* A single-page application is an application that loads a single HTML page and all the necessary assets (such as JavaScript and CSS) required for the application to run. Any interactions with the page or subsequent pages do not require a round trip to the server which means the page is not reloaded.
```
Serrowxd: I would include mention of the virtual DOM, otherwise a very good answer!

Serrowxd:
Components -

I liked your explination, but there are two different types specifically for components - being Functional (doesn't have state) and Class (has state). You can also write them in a more modern fashion, like:

Functional fat-arrow:
`const ComponentName = props =>`

Class Component:
`class ComponentName extends Component`

You would of course need to import Component in your React import as `import React, { Component } from 'react';`

I like it! Nicely done, and good use of `map` when creating new todo notes.

PR #173, opened 2018-03-23
Lisa Canini

LisaCee:


Serrowxd:
# Code

Your CSS file was imported in your `index.html`, this works but it's easier and a better practice to import it somewhere like your `index.js` file, most developers don't look in your HTML when working with React applications. You also placed your CSS file in the public folder, it's also a good practice to put it in your source folder or at least at the root where it can be easily tracked.

You could actually put the entirety of `ToDo.js` inside `App.js`, because you're really just importing it then putting an name inside App, this helps condense the divs down and makes the code a little easier to manage.

Your code inside `ToDo.js` looks fantastic, the only thing I would change is using a more modern syntax for your class component, it would look something like `class ToDo extends Component` and then `import React, { Component } from 'react';`. When you're passing it props, you don't need to give it `this.props` because this is handled locally, but your props are already being passed down so you can simply give it `props.todo`.

Inside `ToDoList.js` your code looks great, the onClick inside your Todo isn't targeting anything so that could be removed - otherwise it looks great.

# Answers

Your answers looked great, but I would have loved to see a little more in-depth on some concepts. It looks like you've got a good idea of what stuff does, though you could have given some bits about things like the Virtual DOM in answer 1. or different types of bundlers on question 3. - webpack for example.

Your definition of State vs Props is semi-correct, a good way of looking;

- Props are passed to children from their parents and cannot be modified by the chidlren, aka they are `readonly`. They are pretty flexible, so you can use them in Functional as well as Class components.

- State is managed locally by the component through Lifecycles, which you utilized very well in your project. State can be modified and read by the component.

Overall I liked what you did, I personally struggled a lot with React when I was going through this section of the course and I know this is an old review, but if you're still in-need of a bit of assistance with React I'll gladly offer you any and all resources at my disposal.

Nicely done!

- Kevin Jolley **Lambda PM**

PR #178, opened 2018-04-03
Jacob Lyerla

JacobLeonLyerla:


DonnieU:
*App.js*: Proper use of class App w/ constructor, super, and needed methods. Methods have descriptive names. Good use of objects and their key/value pairs for .setState. Good design choice of using style.textDectoration to get a line-through. Well laid out render() method including use of spread operators, TodoList component, options for input, and a button.

*TodoList.js*: Good design choice to have modularity by moving a component to its respective file. Proper use of props, .map, and creating a unique key by using todo+index.

*AddLine.js*: Great addition of adding a class AddLine. Looks like you were trying to make the App more modular with this but it isn't used anywhere yet.

Would like to see more commits with appropriate messages. When moving components out to separate files, be sure to create a `components` directory and place the files in that directory.

Overall a very nice job. Keep up the great work!
Rating (0-3): 2

PR #182, opened 2018-04-03
React - Todo

faulkenfett11:
created out first parts of the todo list, makes input and submit button add new task to list


DonnieU:
*index.js*: Good use of class App w/ constructor, super, and succinct methods. Nice use of objects for adding and submitting tasks. Good use of ternary operator and textDecoration in your `strike` method. Your render method is succinct with good use of TaskList component, spread operator, and toggleClick. Input options are of good design choice.

*App.js*: Proper use of callback function using props and .map.

Ideally, and for future reference, what you have in your `App.js` should be in a new file named something like, `ToDoList.js`. Everything in your `index.js` should be in your `App.js`. And your `index.js` shouldn't have been modified. DM me if you have any questions regarding this.

But code base is good overall and the only ding is your file structure. Keep up the great work!

Rating (0-3): 2

PR #184, opened 2018-04-03
React-Todo -worked with kaitlyn --Justin Mathew

jekm321:


justinborek:
## What Went Well:

#### Nice work! Glad to see React doesn't seem to be slowing you down! You were able to complete the main goals, and it's awesome to see you playing with more advanced stuff when you finish. In a lot of cases, students finish an kind of stop working, it's a great habit that you keep pushing. Great commits as well!

## Areas For Improvement:

#### None!

## Questions:

#### What did you end up refactoring in the last commit?

## Rating (1-3):

#### 3

jekm321:
i further delegated components down to task, and i changed the removal feature from using style.display= 'none' into actual removal from this.state, i also changed the toggle of line-through from a css on click toggle to a complete=true/false basis, and i added the local storage save feature for the list and added a save feature so that the list is 100% save-able because the memory didnt change after clearing tasks, i tried to call presist from clearToDos but unless pressed twice it would not truly change local storage.... so save button xD these were all in a single refractor but i split the commits up at points i knew was working to allow rollbacks in case i mess up

PR #185, opened 2018-04-03
React-Todo

rhogan29:
progress on components

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "react-i-new",
+  "name": "react-i",
   "version": "0.1.0",
   "lockfileVersion": 1,
   "requires": true,
--- a/src/App.js
+++ b/src/App.js
@@ -1,9 +1,10 @@
 import React from 'react';
+import toDoList from './Components/toDoList';
+import completed from './Components/completed';
 
 const App = () => (
   <div>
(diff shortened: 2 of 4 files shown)


justinborek:
## What Went Well:

#### Nice work! You were super concerned for no reason! I figured out your error and with only deleting one line, your code works perfectly and passes all the basic requirements. Nice job playing with the styling as well! Great work!

## Areas For Improvement:

#### You were running into an issue with importing CSS `./src/App.js
Module not found: Can't resolve './toDoList.css' in 'C:\Users\AndreaH\Desktop\React-Todo\src'`. So, if you look at that error, the issue is with something in `App.js`. Let's look at the code 
```
import React from 'react';
import ToDoList from './Components/toDoList';
import completed from './Components/completed';
import './App.css';
import './toDoList.css';

const App = () => (
  <div>
    <ToDoList />
  </div>
);

export default App;
```
#### So, two things here. First, because your `toDoList.css` is in the components folder and the `app.js` is in the src folder, you would need to import it with `import './Components/toDoList.css';`. However, because this component isn't the target of that css file, you actually don't need to import it in this component. If you delete that line it works perfectly.

## Questions:

#### Where did your commits go?! 

## Rating (1-3):

#### 2

PR #189, opened 2018-04-03
Eric Gomez React-ToDo

EricGomez101:


DonnieU:
*App.js*: Proper use of class App w/ constructor, super, and needed methods. Methods have descriptive names. Code is easy to read and concise. Proper use of objects and their key/value pairs for .setState. Well laid out render() method including use of spread operators, variables, and "this".

*./components/input.js*: Good design choice in moving component to its own respective file. Nice use of props and options for input.

*./components/list.js*: Another good design choice in separating out components into their respective files as this adds great modularity. Good use of props, .map, and creating a unique key by using listitem + index.

Still would like to see more commits w/ descriptive commit messages.

Great job overall especially with the use of a proper file structure by using a components directory and creating modularity.

Rating (0-3): 2.5

PR #196, opened 2018-04-03
Morales.ReactToDo

Davidless:


Davidless:
This one was tough. I didn't know where, didn't know where things went. Messed up a lot of imports. Got frustrated and lifted a bit from a partners assignment. Still did not get everything passing. He apparently did, but I don't even know what he was writing so attempting to debug was a nightmare. Will likely spend some time with React over the weekend because it looks fairly straightforward as far as a the API goes.

PR #201, opened 2018-04-03
Cruise CS9

CruiseBrantley:


src/App.js
```diff
@@ -1,10 +1,57 @@
...
-  <div>
-    <h2>Todo App</h2>
+const styles = {
+  fontFamily: "sans-serif",
+  textAlign: "center"
+};
 
-  </div>
-);
+// I'll apply this line when I figure out how...
```
kiadorii: You are in the right direction!

PR #203, opened 2018-04-03
Tylar Pierson

tylarpierson:


src/Components/Todo.js
```diff
@@ -0,0 +1,30 @@
...
+    super(props);
+    this.state = {
+      clicked: false
+    };
+  }
+
+  handleClick = () => {
+    const clicked = this.state.clicked;
+    this.setState({
+      clicked: !clicked
```
kiadorii: I'm not 100% certain what is going on here but I think this is where you have a bug.

PR #222, opened 2018-04-04
Anthony-Calderaro's Todo App

Anthony-Calderaro:


DonnieU:
*App.js*: Proper setup of class App w/ constructor, super, and needed methods. Good use of objects and key/value pairs for setState. Succinct code base for render method including use of TodoList component and options for input. Nice addition of a reset button.

*TodoList.js*: Proper use of callback function with props and .map along with creating a unique key by using task + index.

*Todo.js*: Good modularity by having `addTask` in it's own file and function. Proper use of props.text here.

*TaskComplete.js*: Another good example of modularity by having `complete` in its own file and function.

*index.css*: Smart design choice to move styling out to its respective file.

Would still like to see more commits w/ good commit messages.

Very good job overall! Keep up the great work!

Rating (0-3): 2.5

PR #224, opened 2018-04-04
todo pull

kyle-kauzlarich:


DonnieU:
*App.js*: Proper build of class App along w/ constructor and methods to handle adding and submitting a new to-do. Good use of objects for setState. Render method uses ToDoList component appropriately and the same for your input along with an onClick button.

*list.js*: Proper use of callback function using props for your ToDoList. Good design choice creating a unique key for each item.

Overall, very good work with ample commits with great commit messages.

Rating (0-3): 2.5

PR #233, opened 2018-05-01
Andrew Kolumbic

AKolumbic:


src/components/TaskList.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+// import XdOutList from "./XdOutList.js";
+
+const TaskList = props => {
+    return(
+        <ul>
+            {props.tasks.map(task =>{return <li key="task">{task}</li>})}
```
etisdew: Good job demonstrating how to generate bulk elements. :+1:

etisdew: The keys must be unique for each unfortunately.

src/components/ToDoList.js
```diff
@@ -0,0 +1,39 @@
...
+            taskList:[],
+            inputValue:''
+          };
+    }
+    handleChange = event => {
+       this.setState(
+           { inputValue: event.target.value }
+       )
+    }
+    onClick = event => {
```
etisdew: We should check that inputValue is not empty here.

PR #240, opened 2018-05-01
Colin-Preston, Mikeal,Lavell

Colin-Preston:


src/components/todo.js
```diff
@@ -0,0 +1,49 @@
...
+        this.setState({ todos: todos, todo: "" });
+    };
+
+
+
+    render() {
+        return (
+        <div>
+         Todo 
+          <todos todo={this.state.todo} />
```
designerexpert: In this line the class todo does not exist. Not sure what you intended to implement here, but my assumption is that you intended to display the this.state.todo as text inside a div. To accomplish this try this lines:

```<div>{this.state.todo}</div>```

File: src/components/todo.js [hunk 2]
```diff
@@ -0,0 +1,49 @@
...
+    };
+
+
+
+    render() {
+        return (
+        <div>
+         Todo 
+          <todos todo={this.state.todo} />   
+            <input name="todos" onChange={this.nameReciever} value={this.state.todos} placeholder="React palyground" />
```
designerexpert: The input and button belong inside the TodoList class, and not inside the Todo. Inside the TodoList you should instead pass the information that the user inputs as props to each Todo being created.

PR #248, opened 2018-05-01
Brandon Benefield

bbenefield89:


src/App.js
```diff
@@ -1,10 +1,79 @@
...
+    this.state = {
+      inputTextValue: '',
+      todoListTasks: [],
+    };
+  }
+
+  // submit new task
+  handleAddTask = () => {
+    if (!this.state.inputTextValue) {
+      alert('No Task!');
```
etisdew: Excellent work preventing a corner-case. :+1:

src/App.js
```diff
@@ -1,10 +1,79 @@
...
+  }
+
+  // handle task completion (remove task)
+  handleTaskCompletion = e => {
+    const taskID = e.target.id;
+    const stateArr = [...this.state.todoListTasks];
+    
+    for (let i = 0; i < stateArr.length; i++) {
+      if ('' + stateArr[i].ID === taskID) {
+        stateArr.pop();
```
etisdew: This seems wrong... yes, this will only remove the last most item on the list regardless of the target.
Perhaps simply setting display: none; on stateArr[i] would suffice.

etisdew: This would require some kind of handler, perhaps another field in the state array entry properties.

etisdew:
Excellent job telling a story with your commit history. :100:

PR #252, opened 2018-05-01
Jorge Valdivia PR 

emortalix:
Have issues getting the yarn installed and running whenever you get a chance let me know how you were able to get it going

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,4 +1,14 @@
 import React from 'react';
+import TodoList from './components/TodoList'
+
+// Create App Class that extends component
+// State will be holding an empty string for each item entered
+// Empty Array that we will push individual objects such as text content and completed Boolean (T/F)
+// Todo Handler that will hold the typing event
+// Add submit event <button>
+// onComplete Handler
+// Clear Handler
+// Render React Statement
 
 const App = () => (
   <div>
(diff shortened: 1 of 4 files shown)


src/components/Todo.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+const styles = {
+    textDecoration: 'italic'
+};
+const Todo = props => {
+    return (
+        <div style={props.todo.completed ? styles : null}
```
designerexpert: The prop 'todo.completed' does not exist, nor is it passed in upon creation of each todo inside the TodoList class. If you want to have access to this function, you need to pass it when you are maping over ```this.state.todos``` inside the TodoList, I suggest console.logging the props of TodoList so you know how to access the function passed down from App.

src/App.js
```diff
@@ -1,10 +1,77 @@
...
+
+  submitTodo = event => {
+    event.preventDefault();
+    const { todos } = this.state;
+    const myTodo = { text: this.state.todo, completed: false };
+    todos.push(myTodo);
+    this.setState({ todos, todo: ''});
+  };
+
+  completedTodo = childTodo => {
```
designerexpert: Why not handle this inside the TodoList class instead of the App?

PR #254, opened 2018-05-01
Erinc Emer

erinc35:


designerexpert:
# Comments
* I like that the Todo-List worked as expected, implementation of the StrikeOut class was nice. You demonstrate a good understanding of the concepts taught during lecture. 💯 
* Suggestion: Instead of handing the TodoList form, buttons and methods. Perhaps they can be handled inside the TodoList as a class. That way they are completely modular, and can be used inside any other application if needed.

PR #279, opened 2018-05-02
april martinez

april7229:


src/components/component.js
```diff
@@ -1,2 +1,40 @@
 // your components will all go in this `component` directory.
 // feel free to change this component.js into TodoList.js
+import React, { Component } from 'react';
+
+
+class TodoList extends Component {
```
enitchals: It looks like this is identical or nearly identical to the `ToDoList` component. Did you add this as a reference?

PR #287, opened 2018-06-13
Raymond Garcia, React Todo

Raymondgrc:


tysonwilliams:
@Raymondgrc Good job! Btw, for git commit messages, it's usually common convention to use action verbs (to do verbs), like a set of instructions to do to replicate the code.

For example, some of your commit messages could be converted to:
**Setup** project
**Add** in input form and a sample item
**Add** state
**Display** list content, error with displayed content
etc.

😃

PR #340, opened 2018-06-20
Fashoky

Fashoky:
Will add my first name from now on in my PRs!

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,12 +1,11 @@
 import React from "react";
 import ReactDom from "react-dom";
-
 class App extends React.Component {
   constructor() {
     super();
     this.state = {
       greeting: "Todo List: MVP",
-      Todo: ["ToDo"]
+      Todo: [""]
     };
   }
   // you will need a place to store your state in this component.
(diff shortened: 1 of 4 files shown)


bdurb:
@Fashoky Thank you for the PR.  In the future if you could include your first name in the title, that would be helpful.  Thanks!

src/App.js
```diff
@@ -1,16 +1,32 @@
...
-        <h2>Welcome to your Todo App!</h2>
+        <h2>Todo List: MVP</h2>
+        <h1>Learn setstate()</h1>
+        <h1> Style my Todo list</h1>
       </div>
     );
   }
 }
 
+changeGreetingHandler = event => {
```
bdurb: @Fashoky you really don't need to be worried about the greeting handler for this project.  You should be focused on hard coding an array of state objects, and then passing state down as props to your TodoList.  Similar to how bands were passed down in the example code.

bdurb:
@Fashoky I think you may be focusing on the wrong things for this project.  Lets make sure we can get todos to render via props passed to components that we set up.

src/App.js
```diff
@@ -1,16 +1,90 @@
-import React from 'react';
-
+import React from "react";
+import ReactDom from "react-dom";
```
bdurb: no need to import react dom here, its imported in the root of the project index.js

bdurb:
@Fashoky looks like you're moving in the right direction with the project.

PR #343, opened 2018-06-20
Tom Camilleri Todo App

tommycam:


bdurb:
@tommycam thanks for the PR.  Looks like you're off to a good start.

src/App.js
```diff
@@ -1,23 +1,39 @@
...
         "butter"],
       item: ""
-    }
+    };
+  }
+  trackInputHandler = event => {
+    this.setState({item: event.target.value});
+    console.log(event.target.value);
+  }
+  addItem = () => {
```
bdurb: Preventing the default behavior here should fix your refresh issue.

bdurb:
Todo list is looking good.  You will need to switch from an array, to an array of objects to add some of the needed functionality

PR #347, opened 2018-06-20
Kelley Henican - React-Todo

khenic1:


src/components/TodoComponents/TodoList.js
```diff
@@ -1,2 +1,16 @@
+// import App from '../../App.js';
+import React from 'react';
+
+
 // your components will all go in this `component` directory.
 // feel free to change this component.js into TodoList.js
+
+class TodoList extends App {
```
bdurb: @khenic1 the TodoList does not need to be a Class component.  All components outside of App should be stateless functional components.

bdurb:
@khenic1 thank you for the PR.  let me know if you need any assistance.

PR #473, opened 2018-09-25
Colin Dismuke - React-Todo

cpdis:
☑️ @John_Spraul
☑️ @John_Spraul
☑️ @John_Spraul

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -37,6 +37,7 @@ addToDo = (newTask, id) => {
 
   // Set the current state with the new todo array
   this.setState({todos})
+  localStorage.setItem("todos", JSON.stringify(todos));
 }
 
 // Need a handler that takes the input from the form and creates new todo
@@ -81,6 +82,7 @@ toggleComplete = (element) => {
 
      // Set the current state with the new todo array
     this.setState({todos})
+    localStorage.setItem("todos", JSON.stringify(todos));
 }
 
 // Need a function that removes completed todos
@@ -108,14 +110,37 @@ removeComplete = (element) => {
     // Set the current state with the new todo array
     todos = filtered;
     this.setState({todos})
+    localStorage.setItem("todos", JSON.stringify(todos));
+}
 
+hydrateStateWithLocalStorage() {
+  // Iterate through all items in state
+  // This was extremely helpful: https://hackernoon.com/how-to-take-advantage-of-local-storage-in-your-react-projects-a895f2b2d3f2
+  for (let key in this.state) {
+    // Check to see if the key exists in localStorage. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty
+    if (localStorage.hasOwnProperty(key)) {
+      let value = localStorage.getItem(key);
+
+      // Parse the localStorage string and setState
+      try {
+        value = JSON.parse(value);
+        this.setState({ [key]: value })
+      } catch (e) {
+        this.setState({ [key]: value })
+      }
+    }
+  }
+}
+
+componentDidMount() {
+  this.hydrateStateWithLocalStorage();
 }
 
   render() {
     return (
       <div>
         <h1>ToDo List: MVP</h1>
-        <TodoList key={this.state.todos.id} list={this.state.todos} toggleComplete={this.toggleComplete} />
+        <TodoList key={this.state.todos.id} todos={this.state.todos} toggleComplete={this.toggleComplete} />
         <TodoForm submit={this.addToDoHandler} removeComplete={this.removeComplete}/>
       </div>
     );
--- a/src/App.js
+++ b/src/App.js
@@ -138,8 +138,8 @@ componentDidMount() {
 
   render() {
     return (
-      <div>
-        <h1>ToDo List: MVP</h1>
+      <div className="todo-container">
+        <h1>ToDo List: Stretch</h1>
         <TodoList key={this.state.todos.id} todos={this.state.todos} toggleComplete={this.toggleComplete} />
         <TodoForm submit={this.addToDoHandler} removeComplete={this.removeComplete}/>
       </div>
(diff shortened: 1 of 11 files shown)


cpdis:
[React Todo App on Netlify 🌎 ](https://colin-react-todo.netlify.com/)

src/App.js
```diff
@@ -1,16 +1,158 @@
...
+addToDo = (newTask, id) => {
+  // Create a new todo object based on a given task and id
+  const todo = {
+    task: `${newTask}`,
+    id: id,
+    completed: false,
+  }
+
+  // Get the todo array and add the new todo to the end using push()
+  const todos = this.state.todos;
```
John-Spraul: this line should be `const todos = this.state.todos.slice()`, so you make a copy of the current todos in state, but don't actually modify it directly

src/App.js
```diff
@@ -1,16 +1,158 @@
...
+}
+
+// Need a handler that takes the input from the form and creates new todo
+addToDoHandler = (element) => {
+
+  // This prevents the page from being refreshed, otherwise you lose everything when you submit
+  element.preventDefault();
+
+  // Get the input from the node immediately preceding the current one. See: https://developer.mozilla.org/en-US/docs/Web/API/Node/previousSibling
+  const input = element.target.previousSibling.value;
```
John-Spraul: interesting way of solving this problem

Go-to React approach is to hold the input value in state combined with a handling function that changes the value as the form/input is filled out

src/App.js
```diff
@@ -1,16 +1,158 @@
...
+  
+  // Set the current date as the id for the todo
+  const id = Date.now();
```
John-Spraul: easy enough to assign this in the `addToDo` function, then you don't have to worry about passing in an second argument

src/App.js
```diff
@@ -1,16 +1,158 @@
...
+
+  // Return the newly created object and set the state with it using addToDo function
+  return this.addToDo(input, id)
```
John-Spraul: why return if `addToDo` doesn't return anything?

src/App.js
```diff
@@ -1,16 +1,158 @@
...
+    
+    // Get the todo array so that the completed state of the current todo can be toggled between true and false.
+    // Iterate through using .map
+    const todos = this.state.todos.map(todo => {
+
+      // Check if current todo matches the todo id
+      // Convert current.id to a number since it's currently a string
+      if (Number(current.id) === todo.id) {
+        // Add a class that strikes through the word
+        current.classList.toggle('completed');
```
John-Spraul: best to leave the class toggling stuff to React instead of using `classList.toggle`

src/App.js
```diff
@@ -1,16 +1,158 @@
...
+
+        // Sets completed to whatever the opposite value is
+        todo.completed = !todo.completed;
+      }
+
+      // Just return the todo if it doesn't match the current target
```
John-Spraul: this comment is misleading, because the function returns either way

src/App.js
```diff
@@ -1,16 +1,158 @@
...
+     // Set the current state with the new todo array
+    this.setState({todos})
+    localStorage.setItem("todos", JSON.stringify(todos));
+}
+
+// Need a function that removes completed todos
+removeComplete = (element) => {
+
+    // This prevents the page from being refreshed, otherwise you lose everything when you submit
+    element.preventDefault();
```
John-Spraul: `.preventDefault` is a method for events, so element might be the wrong placeholder choice

src/App.js
```diff
@@ -1,16 +1,158 @@
...
+
+    // Create an empty array to hold incomplete items
+    let filtered = [];
+
+    // Get all of the todos with the completed attribute set to false
+    // Use something similar to toggleComplete but use const todos = this.state.todos.filter instead
+    let todos = this.state.todos.filter(todo => {
```
John-Spraul: this isn't the right way to use the `filter` method; you're using it as a `for` loop

src/App.js
```diff
@@ -1,16 +1,158 @@
...
+      return filtered;
+    });
+
+    // Set the current state with the new todo array
+    todos = filtered;
+    this.setState({todos})
+    localStorage.setItem("todos", JSON.stringify(todos));
+}
+
+hydrateStateWithLocalStorage() {
```
John-Spraul: 💧

src/App.js
```diff
@@ -1,16 +1,158 @@
...
+  // Iterate through all items in state
+  // This was extremely helpful: https://hackernoon.com/how-to-take-advantage-of-local-storage-in-your-react-projects-a895f2b2d3f2
+  for (let key in this.state) {
```
John-Spraul: what else are you storing in localStorage other than the todos array?

src/App.js
```diff
@@ -1,16 +1,158 @@
...
+    // Check to see if the key exists in localStorage. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty
+    if (localStorage.hasOwnProperty(key)) {
+      let value = localStorage.getItem(key);
+
+      // Parse the localStorage string and setState
+      try {
+        value = JSON.parse(value);
+        this.setState({ [key]: value })
+      } catch (e) {
```
John-Spraul: what is this `catch` supposed to do?

src/components/TodoComponents/TodoList.js
```diff
@@ -1,2 +1,20 @@
...
+import Todo from './Todo';
+
+const TodoList = (props) => {
+    return (
+      <div className="list-container">
+          {/* Takes as input (from props) each of the todos in state, iterates through (using .map), and renders them onscreen */}
+          {props.todos.map(todo => {
+            return (
+              <div>
+                <Todo key={todo.id} id={todo.id} task={todo.task} onClick={props.toggleComplete} completed={todo.completed} />
```
John-Spraul: you're passing down parts of `todo` as props individually, why not just pass down the whole `todo` object?

PR #474, opened 2018-09-25
Rob Salzberg - React-Todo

robsalzberg:
@juliankohlman

**Day 1 Challenges**

- [x] Your todo list should display a list of todos, an input field, a submit button, and a clear all button.
Be sure to use the given files for building out these components.

- [x]  App.js will hold all the data needed for this project. It will also be the container for your Todo Components.
      All of your application data will be stored here on <App />.
     All of your handler functions should live here on <App />.

- [x] TodoList.js receives your Todos array and iterates over the list generating a new <Todo /> for each element in the array.

- [x] Todo.js is a component that takes in the todo data and displays the task to the screen.

- [x] TodoForm.js will hold your input field and your Add Todo and Clear Completed buttons.

- [x] Your  input field should take in user input, and allow a user to press Enter or click on the Submit Button to add a todo to your list.

- [x] Once  a todo is submitted, the Todo List should re-render and show the added todo.

**Day 2 Challenges**

- [x] Add  the functionality to toggle your todo's completed flag from false to true.

- [x] Once  a todo is completed, be sure to demonstrate to the user that the todo is completed by adding a line-through style property if the completed flag is true.

- [x] Add the ability to remove any todos that you have completed. .filter will be your best friend here.

- [x]  When a user clicks on the Clear Completed button call your handler function that will filter out any todos that have the completed flag toggled to true.

- [x] Now is the time to style Take your time to make this an app that you can be proud of.

**Stretch Problems**

- [ ] Persist  your data in window.localStorage() hint: you may have to pass your data to a stringifier to get it to live inside the localStorage() of the browser. This will cause it to persist past the page refresh.

- [ ] Search  Functionality Add a input bar that allows you to search through your tasks and only show the ones that match the search input.

- [ ] Hosting  Create a Netlify Account and follow the tutorial on how to host your shiny new todo app on the world wide web.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,13 +1,36 @@
-import React from 'react';
+import React from "react";
+import TodoList from "./components/TodoComponents/TodoList.js";
+import TodoForm from "./components/TodoComponents/TodoForm.js";
+import './components/TodoComponents/Todo.css';
 
 class App extends React.Component {
-  // you will need a place to store your state in this component.
-  // design `App` to be the parent component of your application.
-  // this component is going to take care of state, and any change handlers you need to work with your state
+  constructor() {
+    super();
+    this.state = {
+      todoData: []
+    };
+  }
+
+  addTask = task => {
+    const newTask = {
+      task: task,
+      id: Date.now(),
+      isCompleted: false
+    };
+
+    this.setState({
+      todoData: [...this.state.todoData, newTask]
+    });
+
+    console.log(this.state.todoData);
+  };
+
   render() {
     return (
       <div>
-        <h2>Welcome to your Todo App!</h2>
+        <h1>Todo List</h1>
+        <TodoList list={this.state.todoData} />
+        <TodoForm addTask={this.addTask} />
       </div>
     );
   }
--- a/src/App.js
+++ b/src/App.js
@@ -1,36 +1,60 @@
-import React from "react";
-import TodoList from "./components/TodoComponents/TodoList.js";
-import TodoForm from "./components/TodoComponents/TodoForm.js";
+import React from 'react';
+import TodoList from './components/TodoComponents/TodoList';
+import TodoForm from './components/TodoComponents/TodoForm';
 import './components/TodoComponents/Todo.css';
 
 class App extends React.Component {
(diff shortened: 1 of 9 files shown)


src/App.js
```diff
@@ -1,36 +1,60 @@
...
+    constructor() {
+      super();
+      this.state = {
+        tasks: [],
+        holder: "",
+      }
+    }
+    toggleClass = (event, id) => {
+      event.target.classList.toggle('completed');
+      this.setState({
```
juliankohlman: Overall, your project is looking good Rob. Take a look at the snippet below and the comments 
```javascript
let tasks = this.state.tasks.slice(); // copy of todos

    tasks = tasks.map(task => {
      if (task.id === id) {
        task.completed = !task.completed; // only update the prop that has been changed
        return task; // return the updated task
      } else {
        return task;
      }
    });

    this.setState({ tasks }); // setting state w/new tasks array created above instead of from within the setState method. 
```
We should try and avoid placing too much logic inside `setState` itself. Check out this [Article on setState](https://itnext.io/react-setstate-usage-and-gotchas-ac10b4e03d60)

src/App.js
```diff
@@ -1,36 +1,60 @@
...
+        tasks: this.state.tasks.map(task => {
+          if (task.id === id) {
+            return {
+              task: task.task,
```
juliankohlman: Like I mentioned above you could stick with the pattern you're using and condense the code some with something like below:
``` javascript
if (task.id === id) {
          task.completed = !task.completed;
          return task;
        } else {
          return task;
        }
```
In the code above we're not overwriting the task and returning but just updating the relevant task property and then returning the task.

src/components/TodoComponents/Todo.js
```diff
@@ -1,5 +1,9 @@
-import React from "react";
-const Todo = props => {
-  return <li>{props.todoItem}</li>;
-};
-export default Todo;
+import React from 'react';
+
+const Todo = (props) => {
+    console.log(props);
```
juliankohlman: 👍 for using `console.log` to poke around and see what's happening with `props`.

PR #480, opened 2018-09-25
Michael Littleton - React Todo

mglittleton:


eddygonzalez9708:
Great job at completing MVP. Your submit handler method that adds todo items should prevent empty strings from being added to the todo list.  You can write a condition to check if the input field is empty before triggering that set state method. On the other hand, I like how your todo list container's height changes when adding todo items and I like how you implemented array loop methods to find a specific item in the todo list so you can change its completed property. However, as your list continues to grow it may not be efficient to use an array loop method such as filter or map. Instead, you could have provided an index to the index attribute when mapping over each todo item in your Todo List component like: 

```
 {props.list.map((item, index) => (
  <Todo index={index} key={item.id} tag={item.id} complete={item.complete} task={item.task} completeHandler={props.completeHandler}/>
))}
```

Then, in the Todo component you pass down the index to the complete handler. 

```<li className={`normal-todo ${props.complete ? "strike-through-todo" : null}`} onClick={() => props.completeHandler(props.index)}>```

Then, when the todo item is clicked on you can implement your item completed handler to toggle complete like this: 

```
completeHandler = (index) => {
const arr = [...this.state.list]
this.setState({
list: arr
)}
}
```

PR #484, opened 2018-09-25
Lee Lichtsinn - React-Todo

leelichtsinn:
added constructor, super and state
@47analogy

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,8 @@
 import React from 'react';
 
+import TodoList from './components/TodoComponents/TodoList.js';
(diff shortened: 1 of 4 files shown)


leelichtsinn:
Added TodoList
Added Todo
Added sample state to App.js
Added TodoForm
@47analogy

47analogy:
Basic static structure of the app looks good. To add the functionality, you would have to add an event handler to the button (to wire up everything), write your function to do something (i.e. add a todo) and update the state (using setState) to re-render. Please reach out when you can so that we can go over this.

PR #486, opened 2018-09-25
Kurt Van Etten

kurtve:
@gakko1  Initial commit


kurtve:
Turned out that I misspelled 'TodoForm' in the class declaration. :)
So, at least the form appears now.  I will flesh out the functionality tomorrow.

kurtve:
So evidently HTML text input fields have an attribute called 'placeholder'.  Who knew! :)
Anyway, even though it's extra code cluttering the app, I kind of like my functionality better than what the browser does by default, so I'll leave things alone.

PR #490, opened 2018-09-25
Eric Fleming - React-Todo

efleming111:
React-Todo @gakko1

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -23,10 +23,21 @@ class App extends React.Component {
   }
 
   componentDidMount(){
-    // Rewrite set variable to JSON then set headerText then setState
-    if(JSON.parse(localStorage.getItem("data")) !== null){
+    const data = JSON.parse(localStorage.getItem("data"));
+
+    let index = Number.parseInt(data.length / 3, 10);
+    index = index >= this.headers.length ? this.headers.length - 1 : index;
+
+    if(data !== null){
       this.setState({
-        todoData: JSON.parse(localStorage.getItem("data"))
+        todoData: JSON.parse(localStorage.getItem("data")),
+        headerText: this.headers[index]
+      })
+    }
+
+    else{
+      this.setState({
+        headerText: this.headers[index]
       })
     }
   }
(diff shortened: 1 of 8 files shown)


src/components/TodoComponents/SearchForm.js
```diff
@@ -0,0 +1,42 @@
...
+        }
+    }
+
+    handleInput = (event)=>{
+        this.setState({
+            searchKey: event.target.value
+        })
+    }
+
+    hideTask = (event)=>{
```
gakko1: Why are these on here twice? You can pass them from App.js and pass a parameter from this component's state into the props method then you would just need an input handler and searchKey state.

src/components/TodoComponents/TodoForm.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    handleClearCompleted = (event)=>{
+        event.preventDefault();
+        this.props.clearCompleted();
+    }
+    
+    render(){
+        return(
+            <form className="todo-form">
```
gakko1: You can simplify this a little bit with an `onSubmit` property on your form element, calling a submit handler that takes care of both clicking the submit button or pressing enter. Then you'd only need an `onClick` property on the clear button to differentiate the two.

src/App.js
```diff
@@ -1,13 +1,133 @@
...
+      index = 0;
+    }
+
+    this.setState({
+      todoData: data,
+      headerText: this.headers[index]
+    });
+  }
+
+  hideTask = (searchKey)=>{
```
gakko1: This along with the showAll method are brilliant yet simple, good implementation of a search system.

src/App.js
```diff
@@ -1,13 +1,133 @@
...
+      id: Date.now(),
+      isCompleted: false,
+      isHidden: false
+    };
+
+    let index = Number.parseInt(this.state.todoData.length / 3, 10);
+    index = index >= this.headers.length ? this.headers.length - 1 : index;
+
+    this.setState({
+      todoData: [...this.state.todoData, newTask],
```
gakko1: Nice use of the spread operator

PR #491, opened 2018-09-25
Ryan McLaughlin

rlmclaughlin:
#### Day 1 Challenges
-[x] Your todo list should display a list of todos, an input field, a submit button, and a clear all button.
-[x] Be sure to use the given files for building out these components.
-[x] `<App />` will hold all the data needed for this project. It will also be the container for your Todo Components.
  -[x] All of your application data will be stored here on `<App />`.
  -[x] All of your `handler` functions should live here on `<App />`.
-[x] `<TodoList />` receives your Todos array and iterates over the list generating a new `<Todo />` for each element in the array.
-[x] `<Todo />` is a component that takes in the `todo` data and displays the task to the screen.
-[x] `<TodoForm>` will hold your input field and your `Add Todo` and `Clear Completed` buttons.
  -[x] Your input field should take in user input, and allow a user to press `Enter` or click on the `Submit Button` to add a todo to your list.
  -[x] Once a todo is submitted, the Todo List should re-render and show the added todo.

Day 2:
[x] Add the functionality to toggle your todo's completed flag from false to true.
    Once a todo is completed, be sure to demonstrate to the user that the todo is completed by adding a 
    line-through style property if the completed flag is true.
[x] Add the ability to remove any todos that you have completed. .filter will be your best friend here. 
     When a user clicks on the Clear Completed button call your handler function that will filter out any 
     todos that have the completed flag toggled to true.
[x] Now is the time to style Take your time to make this an app that you can be proud of.
    Stretch Pr

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,12 +1,22 @@
 import React from 'react';
 
 class App extends React.Component {
+  constructor(){
+  super(); 
+  
+    this.state = {
+      todos: " ",
+      id: " ", 
+      completed: false 
+    }
+  }
+
   // you will need a place to store your state in this component.
   // design `App` to be the parent component of your application.
   // this component is going to take care of state, and any change handlers you need to work with your state
   render() {
     return (
-      <div>
+      <div className="App">
         <h2>Welcome to your Todo App!</h2>
       </div>
     );
--- a/src/App.js
+++ b/src/App.js
@@ -5,10 +5,17 @@ class App extends React.Component {
   super(); 
   
     this.state = {
-      todos: " ",
-      id: " ", 
-      completed: false 
+      todos: [
+      { list: " ", id: 0, completed: false }
+      ] 
     }
+  }
+
+  handler(){
+    this.setState({
+     task: event.target.value,
+     id: dateNow()
+    })
   }
 
   // you will need a place to store your state in this component.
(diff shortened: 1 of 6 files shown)


eddygonzalez9708:
You did great job with this project. The styling of your application is awesome, but just be wary of items displaying outside of the design as you continue adding items to your todo list. I also like how you implemented array loop methods to find a specific item in the todo list so you can change its completed property. However, as your list continues to grow it may not be efficient to use an array loop method such as filter or map. Instead, you could have provided an index to the index attribute when mapping over each todo item in your Todo List component like:

```{props.todolist.map((newToDo, index) => { return ( <ToDo index={index} key={newToDo.id} value={newToDo.task} textStrike={props.textStrike} id={newToDo.id} completed={newToDo.completed} />) })}```

Then, in the Todo component you pass down the index and assign as an argument to the techSrike method. 

```const Todo = (props) => { return (<li id={props.id} onClick={()=>{props.textStrike(props.index)}} className={ (props.completed ? 'completed' : null)}>{props.value}</li>)}```

Then, when the todo item is clicked on you can implement your techStrike handler to toggle complete like this:

```
textStrike = index => 
{ const arr = [...this.state.todolist] 
this.setState({ todolist: arr})}```

PR #504, opened 2018-09-26
Wes Meeks - React-Todo

wesmeeks88:
[ ] Don't focus on styling yet. We want you to worry about function over form today.
 [ ] Your todo list should display a list of todos, an input field, a submit button, and a clear all button.
Be sure to use the given files for building out these components.
<App /> will hold all the data needed for this project. It will also be the container for your Todo Components.
 [ ] All of your application data will be stored here on <App />.
 [ ] All of your handler functions should live here on <App />.
 [ ] <TodoList /> receives your Todos array and iterates over the list generating a new <Todo /> for each element in the array.
 [ ] <Todo /> is a component that takes in the todo data and displays the task to the screen.
 [ ] <TodoForm> will hold your input field and your Add Todo and Clear Completed buttons.
 [ ] Your input field should take in user input, and allow a user to press Enter or click on the Submit  Button to add a todo to your list.
 [ ] Once a todo is submitted, the Todo List should re-render and show the added todo.
Day 2 Challenges
 [ ] Add the functionality to toggle your todo's completed flag from false to true.
 [ ] Once a todo is completed, be sure to demonstrate to the user that the todo is completed by adding a line-through style property if the completed flag is true.
 [ ] Add the ability to remove any todos that you have completed. .filter will be your best friend here. When a user clicks on the Clear Completed button call your handler function that will filter out any todos that have the completed flag toggled to true.
 [ ] Now is the time to style Take your time to make this an app that you can be proud of.
Stretch Problems
 [ ] Persist your data in window.localStorage() hint: you may have to pass your data to a stringifier to get it to live inside the localStorage() of the browser. This will cause it to persist past the page refresh.

 [ ] Search Functionality Add a input bar that allows you to search through your tasks and only show the ones that match the search input.

 [ ] Hosting Create a Netlify Account and follow the tutorial on how to host your shiny new todo app on the world wide web.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import TodoList from './components/TodoComponents/TodoList';
 
 class App extends React.Component {
   // you will need a place to store your state in this component.
@@ -6,9 +7,7 @@ class App extends React.Component {
   // this component is going to take care of state, and any change handlers you need to work with your state
   render() {
     return (
-      <div>
-        <h2>Welcome to your Todo App!</h2>
-      </div>
+      <TodoList />
     );
   }
 }
--- a/src/components/TodoComponents/Todo.js
+++ b/src/components/TodoComponents/Todo.js
@@ -1,0 +1,12 @@
+import React from 'react';
+import './Todo.css';
+
+const Todo = (props) => {
+  return (
+    <div className="toDoContainer">
+     <h1>{props.todo}</h1>
+     </div>
+  );
+}
+
+export default Todo;
--- a/src/components/TodoComponents/TodoForm.js
+++ b/src/components/TodoComponents/TodoForm.js
@@ -1,0 +1,10 @@
+import React from 'react';
+// import '.Todo.css';
+
+const TodoForm = () => {
+  return (
+    <input type ='text' placeholder='put something in your list!' />
+  );
+}
+
+export default TodoForm;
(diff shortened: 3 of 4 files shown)


eddygonzalez9708:
Great job at completing MVP. Your click handler method that adds todo items should prevent empty strings from being added to the todo list.  You can just code a condition to check if the input field is empty before triggering that set state method. On the other hand, I like how you implemented array loop methods to find a specific item in the todo list so you can change its completed property. However, as your list continues to grow it may not be efficient to use an array loop method such as filter or map. Instead, you could have provided an index to the index attribute when mapping over each todo item in your Todo List component like: 

```
{props.todoList.map((task, index) => {
  return <Todo index={index} itemCompleted={props.itemCompleted} id={task.id} task={task.task} completed={task.completed} />
})}
```

Then, in the Todo component you pass down the index and assign it to the name attribute. 

```<li name={props.index} style={props.completed ? {textDecoration: 'line-through'} : null} class={`todo ${props.task}`} onClick={props.itemCompleted}>{props.task}</li>```

Then, when the todo item is clicked on you can implement your item completed handler to toggle complete like this: 

```
itemCompleted = (event) => {
const arr = [...this.state.todoList] /
const index = Number(event.target.attributes.name.nodeValue) /* Number() changes the string to a number such as '3' to 3 */
this.setState({
todoList: arr
)}
}
```

PR #506, opened 2018-09-26
Jess Harrison React-Todo

jessharrison83:
Day 1 Challenges
- [x] Your todo list should display a list of todos, an input field, a submit button, and a clear all button.
- [x] Be sure to use the given files for building out these components.
- [x] `<App />` will hold all the data needed for this project. It will also be the container for your Todo Components.
  - [x] All of your application data will be stored here on `<App />`.
  - [x] All of your `handler` functions should live here on `<App />`.
- [x] `<TodoList />` receives your Todos array and iterates over the list generating a new `<Todo />` for each element in the array.
- [x] `<Todo />` is a component that takes in the `todo` data and displays the task to the screen.
- [x] `<TodoForm>` will hold your input field and your `Add Todo` and `Clear Completed` buttons.
  - [x] Your input field should take in user input, and allow a user to press `Enter` or click on the `Submit Button` to add a todo to your list.
  - [x] Once a todo is submitted, the Todo List should re-render and show the added todo.


eddygonzalez9708:
You did an awesome job with this project. I like the styling of your application and I like how you implemented array loop methods to find a specific item in the todo list so you can change its completed property. However, as your list continues to grow it may not be efficient to use an array loop method such as filter or map. Instead, you could have provided an index to the name attribute when mapping over each todo item in your Todo List component like: 

{props.tasks.map((task, index) => {
  return ```<Todo name={index} item={task.task} clickHandler={props.clickHandler} />```
})}
 
Then, when the todo item is clicked on you can implement your click handler to toggle complete like this: 

clickHandler = (event) => {
const arr = [...this.state.tasks] /* Spread operator grabs all array items and assigns them to arr */
const index = Number(event.target.name) /* Number() changes the string to a number such as '3' to 3 */
this.setState({
tasks: arr
)}
}

PR #509, opened 2018-09-26
Jesse Anderson - React-Todo

Jesse9009:
@Juliankohlman

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -16,32 +16,54 @@ class App extends React.Component {
     };
   }
 
-  render() {
-    return (
-      <div>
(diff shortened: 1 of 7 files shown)


src/App.js
```diff
@@ -1,13 +1,84 @@
...
+
+  handleInput = event => {
+    this.setState({ textField: event.target.value });
+  };
+
+  toggleComplete = event => {
+    // event.preventDefault();
+    this.setState({
+      todoData: this.state.todoData.map(item => {
+        if (item.id == event.target.id) {
```
juliankohlman: I'd use  `===` just to be on the safe side.

PR #510, opened 2018-09-26
initial PR Michael Trew

mtrew2015:
@juliankohlman

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,23 +1,57 @@
 import React from 'react';
+import TodoList from './components/TodoComponents/TodoList';
+import TodoForm from './components/TodoComponents/TodoForm';
 
 class App extends React.Component {
-	constructor(props) {
-		super(props);
+	constructor() {
+		super();
 		this.state = {
-			items: []
-			//values which we are storing
+			todoList: [
+				{
+					task: "",
+					id: Date.now(),
+					completed: false
+				}
+			],
+			inputText: ''
 		};
 	}
-	// you will need a place to store your state in this component.
-	// design `App` to be the parent component of your application.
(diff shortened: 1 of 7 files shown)


src/App.js
```diff
@@ -1,16 +1,81 @@
...
+			return item.completed === false;
+		});
+
+		this.setState({
+			todoList: newTodoList
+		});
+	};
+	addToList = event => {
+		event.preventDefault();
+		let newArray = this.state.todoList;
```
juliankohlman: Remember to use the `const` keyword until you can't.

src/App.js
```diff
@@ -1,16 +1,81 @@
...
+	changeUserInput = event => {
+		this.setState({ inputText: event.target.value });
+	};
+
+	render() {
+		return (
+			<div>
+				<h1>Michaels Awesome To-Do List</h1>
+				<TodoForm
+					className="todoList"
```
juliankohlman: Nice job, spreading the props across multiple lines. 👍

src/components/TodoComponents/Todo.js
```diff
@@ -0,0 +1,20 @@
...
+// / - `<Todo />` is a component that takes in the`todo` data and displays the task to the screen.
+
+const Todo = props => {
+	return props.list.map((item, id) => (
+		<li
+			onClick={() => props.toggleComplete(item.id)}
+			key={id}
+			id={id}
+			style={{
+				textDecoration: item.completed ? 'line-through' : 'none'
```
juliankohlman: Nice job using inline styles here. Some argue that it's not best practice, but in this case, when you need to apply a style based on some simple logic, this is a very direct route.

PR #511, opened 2018-09-26
Abi Franklin React ToDo

AbiFranklin:
@gakko1

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -7,14 +7,7 @@ class App extends React.Component {
     super(props);
 
     this.state = {
-      todo: [
-        {task: 'Go To Store',
-         id: 123,
-         completed: false},
-        {task: 'Go To Work',
-         id: 124,
-         completed: false},
-      ]
+      todo: []
     }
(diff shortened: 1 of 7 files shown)


src/App.js
```diff
@@ -1,16 +1,101 @@
...
+    super(props);
+
+    this.state = {
+      todo: []
+    }
+;  }
+
+    addItem = (e) => {
+      let temp = {
+        task: e.target.previousSibling.value,
```
gakko1: This is probably better to be done with `setState` and an item in state that holds the value changing dynamically as you type rather than messing with DOM elements.

src/App.js
```diff
@@ -1,16 +1,101 @@
...
+      temparr.forEach(item => {
+        item.completed = false
+      })
+
+      this.setState({
+        todo: temparr
+      })
+
+      console.log(this.state.todo)
+      console.log(temparr)
```
gakko1: Don't forget to remove console logs on final submission

PR #512, opened 2018-09-26
Luis Alvarez: To-do List

TheGameFreak720:
@juliankohlman


src/App.js
```diff
@@ -1,13 +1,96 @@
...
 
 class App extends React.Component {
   // you will need a place to store your state in this component.
   // design `App` to be the parent component of your application.
   // this component is going to take care of state, and any change handlers you need to work with your state
+  constructor() {
+    super();
+    this.state = {
+      userInput: '',
+      backgroundColors: ['#f1c40f', '#e67e22', '#e74c3c', '#1abc9c', '#2ecc71', '#3498db', '#9b59b6'],
```
juliankohlman: Really like this idea of color coding the todos. In the future, you could build this out even more by giving each task a category and tying the colors to those categories. 👍

src/App.js
```diff
@@ -1,13 +1,96 @@
...
+
+  handleSubmit = event => {
+    this.setState({
+      userInput: event.target.value
+    });
+  }
+
+  markDone = (id) => {
+    this.setState({
+      todoList: this.state.todoList.map(todo => {
```
juliankohlman: I would bring all your logic for updating the state out of the `setState` function. This will make your life easier when debugging. By following this pattern your calls to `setState` will look much simpler.

src/App.js
```diff
@@ -1,13 +1,96 @@
...
+        if(todo.id === id && todo.completed === false) {
+          return {
```
juliankohlman: For each todo object, all you need to do here is make the relevant update on the `completed` property then return the todo as you map over them. Could look like the snippet below:
```javascript
if (todo.id === id) { // not checking the completed prop since it's set to false by default
    todo.completed = !todo.completed; // just toggling the completed condition.
    return todo;
}
```

PR #518, opened 2018-09-26
all functionality complete!

CheapCharlie:


47analogy:
To make this project better, try to separate the functionality into the various components via props (keep the state in your main App.js component), according to the ReadMe and lecture on Wednesday night. Also, a way to go about toggling the todos is to initialize the state to complete: false, then create a function that uses the Logical not operator to update the state, then create an appropriate style for toggling in the css. Please reach out with any questions.

CheapCharlie:
I still have to separate into subcomponents. I'll get to that the first moment I have time after todays sprint is completed.

PR #519, opened 2018-09-26
Lorenzo-Evans-React-Todo

LorenzoEvans:
Currently working out how to get the button to push the text from the input field, without the input field pushing it. The button itself is pushing the empty string from state. I _think_ the issue is the logic inside of one or both of my setState methods.


47analogy:
Good habit of commenting your code (especially for learning new concepts), so that when you come back to it later...it's easier to determine what the code is doing. Also can perhaps being more descriptive with the handle names (i.e.handlClickDoneTodo...or something like this instead of handleClick so that it's clear what it does). Good job of adding some creativity when clearing the to do list.

PR #520, opened 2018-09-26
Brandon Moll TODO project

BrandonMoll:
@gakko1


src/App.js
```diff
@@ -1,13 +1,97 @@
...
+            key: todo.key,
+            completed: !todo.completed,
+            todoClass: todo.todoClass === 'incomplete' ? 'complete' : 'incomplete'
+            }
+          } else {
+            return todo
+          }
+      }),   
+    })
+    console.log(this.state.todos)
```
gakko1: Don't forget to remove console logs on final submission

src/App.js
```diff
@@ -1,13 +1,97 @@
...
-  // design `App` to be the parent component of your application.
-  // this component is going to take care of state, and any change handlers you need to work with your state
+  constructor() {
+    super();
+    this.testRef = React.createRef();
+    this.state = {
+      currentInputValue: '',
+      todos : []
+      };
+      window.onbeforeunload = this.saveData;
```
gakko1: This is nice, you could also do it with React lifecycle methods or change it whenever you update with new todos.

src/App.js
```diff
@@ -1,13 +1,97 @@
...
+
+  handleInput = (event) => {
+    this.setState({
+      currentInputValue: event.target.value
+    });
+  }
+
+  handleSubmit = (event) => {
+
+    if (this.state.currentInputValue === '') {
```
gakko1: Good job checking for empty todos, how about the case of multiple spaces for an "empty" todo?

src/App.js
```diff
@@ -1,13 +1,97 @@
...
+      alert('No Todo Item Entered')
+    } else {
+    let newTaskList = this.state.todos.concat({task: this.state.currentInputValue, key: Date.now(), completed: false, todoClass: 'incomplete'})
```
gakko1: You can do this inside the `setState` method by using a rest operator `[...this.state.todos, {new object with changes}]`

src/App.js
```diff
@@ -1,13 +1,97 @@
...
+    this.setState({todos: newTaskList, currentInputValue: ''});
+    console.log(this.state.todos);
+    }
+  }
+
+  onKeyPress = (event) => {
```
gakko1: This is a good work around, you can also use a `form` selector and have an `onSubmit` property that will then work with either clicking the button or pressing enter, while also not needing that method on the button if the button is nested inside the form with proper labeling.

src/App.js
```diff
@@ -1,13 +1,97 @@
...
+    });
+    console.log(this.state.todos)
+  }
+    
   render() {
     return (
-      <div>
-        <h2>Welcome to your Todo App!</h2>
+      <div className="container">
+        <h1 className="header">Tasks Remaining {this.state.todos.length}</h1>
```
gakko1: Nice added feature

PR #535, opened 2018-09-26
Marguel Gutierrez React To Do

Marguelgtz:


src/components/TodoComponents/TodoForm.js
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+const TodoForm = props => {
+    return (
+       <form>
```
gakko1: You can put your submit method on here on an `onSubmit` property to do the same thing as your button plus being able to submit using the enter key.

PR #537, opened 2018-09-26
David Reichardt Todo List

davereichardt:
@raymondgrc

...

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -5,9 +5,20 @@ import TodoList from './components/TodoComponents/TodoList';
 class App extends React.Component {
   constructor(){
     super();
-    this.state={
+    this.state = {
       textInput: '',
-      todos: []
+      todos: [
+        {
+          task: 'Organize Garage',
+          id: 1528817077286,
+          completed: false
+        },
+        {
+          task: 'Bake Cookies',
+          id: 1528817084358,
+          completed: false
+        }
+      ]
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -1,16 +1,48 @@
...
-  // this component is going to take care of state, and any change handlers you need to work with your state
+  constructor(){
+    super();
+    this.state={
+      textInput: '',
+      todos: []
+    };
+  }
+
+  handleInput = (event) => {
```
Raymondgrc: Nice handler!

src/App.js
```diff
@@ -1,16 +1,48 @@
...
+    this.setState({
+      textInput: event.target.value
+    });
+  };
+
+  handleClick = (event) => {
+    event.push();
```
Raymondgrc: So what you may want to do is make an array that is a copy of the todolist, push whatever is in the form to that copy, then setstate to overwrite the todolist with the copy.

src/components/TodoComponents/TodoForm.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+const TodoForm = (props) => {
+    return (
+        <form>
```
Raymondgrc: Nice component, it is well written and easy to see what you're attempting here.

PR #538, opened 2018-09-27
WITANDAY  KYANGA

Witanday:
- []  Day I - Completed Challenge

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,3 +65,6 @@
 - **Search Functionality** Add a input bar that allows you to search through your tasks and only show the ones that match the search input.
 
 - **Hosting** Create a [Netlify Account](https://www.netlify.com/) and follow the tutorial on how to host your shiny new todo app on the world wide web.
+
+
+# Initial Commit
--- a/README.md
+++ b/README.md
@@ -67,4 +67,4 @@
 - **Hosting** Create a [Netlify Account](https://www.netlify.com/) and follow the tutorial on how to host your shiny new todo app on the world wide web.
 
 
-# Initial Commit
+# Initial commit
--- a/src/App.css
+++ b/src/App.css
@@ -1,0 +1,30 @@
+*{
+    padding: 0;
+    margin: 0;
+    box-sizing: border-box;
+    font-size: 62.5%;
+    
+}
+
+body{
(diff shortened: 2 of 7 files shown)


Witanday:
Day 2 Challenges
- [ ]   Add the functionality to toggle your todo's completed flag from false to true.
- [ ]  Once a todo is completed, be sure to demonstrate to the user that the todo is completed by adding a line-through style property if the completed flag is true.
- [ ]  Add the ability to remove any todos that you have completed. .filter will be your best friend here. - [ ]- - [ ] When a user clicks on the Clear Completed button call your handler function that will filter out any todos that have the completed flag toggled to true.
 - [ ]  Now is the time to style Take your time to make this an app that you can be proud of.

eddygonzalez9708:
Great job at completing MVP. There is no need to bind your class methods in React because React does it automatically unless you are using Reactstrap. Also, your search input field should be separate from your add todo input field and your search field should filter based on what is typed in it. If no characters are typed in it, then it should display all the todo items. Here is a video that will help you with implementing a search filter in React: https://www.youtube.com/watch?v=OlVkYnVXPl0. 

On the other hand, I like how you implemented array loop methods to find a specific item in the todo list so you can change its completed property. However, as your list continues to grow it may not be efficient to use an array loop method such as filter or map. Instead, you could have provided an index to the index attribute when mapping over each todo item in your Todo List component like: 

```
<ul>{props.items.map((item, index) =>(
     Todo index={index} toggletodoCompl={props.toggleCompl} key = {item.id} itemkey = {item.id} item 
     = {item.inputTask} DisplayStyle ={item.completed} />
))}
</ul>
```

Then, in the Todo component you pass down the index to the onclick handler. 

```
const Todo = (props) =>{
  return <li className={props.c}  id ={'id' + `${props.itemkey}`} onClick={() => props.toggletodoCompl(props.index)} 
  className={props.DisplayStyle.toString()}>{props.item} </li>
}
```

Then, when the todo item is clicked on you can implement your toggle completed handler to toggle complete like this: 

```
toggleCompleted = (index) => {
const arr = [...this.state.items]
this.setState({
items: arr
)}
}
```

PR #539, opened 2018-09-27
Steve Kutz

stevekutz:
added checkbox & functionality to verify if user tries to add an empty task

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,54 +1,61 @@
 import React from 'react';
-import FormInput from '../src/components/TodoComponents/TodoForm';
-import ListItems from '../src/components/TodoComponents/TodoList';
-
+import TodoList from './components/TodoComponents/TodoList';
+import TodoForm from './components/TodoComponents/TodoForm';
+import './components/TodoComponents/Todo.css';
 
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -1,13 +1,54 @@
...
+  handleSubmit = (event) => {
+    event.preventDefault();
+    const myObj = {
+      task: this.state.input,
+      id: Date.now(),
+      completed: false
+    };
+
+    if(this.state.input === ''){
+      alert('please enter something');
```
clarakosi: I really like this here! Great job making sure you don't have a bunch of empty tasks!

src/components/TodoComponents/Todo.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import './Todo.css'
+
+const Todo = (props) => {
+
+  return (
+    <li>
+      {props.value}
+       <input type = 'checkbox' value  = 'completed'/>
```
clarakosi: You don't need to space this out. Type could be written as such: `type='checkbox'`

PR #541, opened 2018-09-27
Finally able to understand and start the to do list. MVP Day 1 completed. Working on MVP day 2

CJAponte:
@raymondgrc

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,4 +1,7 @@
 import React from 'react';
+import ToDoList from './components/TodoComponents/TodoList';
+import ToDo from './components/TodoComponents/Todo';
+import ToDoForm from './components/TodoComponents/TodoForm';
 
 class App extends React.Component {
   // you will need a place to store your state in this component.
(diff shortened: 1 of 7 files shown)


src/App.js
```diff
@@ -7,7 +10,15 @@ class App extends React.Component {
   render() {
     return (
       <div>
-        <h2>Welcome to your Todo App!</h2>
+        <div>
```
Raymondgrc: I would advise against adding a lot of div tags in your app.js. This is where you want to bring in all your components, so if you really need those divs, why not add them to your components?

src/App.js
```diff
@@ -7,7 +10,15 @@ class App extends React.Component {
...
+          <ToDoList />
+        </div>
+        <div>
+          <ToDo />
```
Raymondgrc: Your ToDoList component should house this component.

src/components/TodoComponents/Todo.css
```diff
@@ -0,0 +1,44 @@
+.todoTitle{
```
Raymondgrc: I appreciate that you spent time styling! Don't let those CSS skills slip up.

src/components/TodoComponents/Todo.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import './Todo.css';
+
+class ToDo extends React.Component{
+    constructor(){
+        super();
+
+        this.state = {
```
Raymondgrc: I would highly recommend having this state info inside the App component, subsequently you would pass the state from the App component to the TodoList component, with that info you would make a new ToDo with each element inside the array.

PR #546, opened 2018-09-29
React Todo - Henry Moland PR

henrymoland:


47analogy:
App.cs is not being importing into app.js. You can fix it by changing it the path to: import './app.css'; Otherwise, basic MVP is complete and functions as expected. Your code is readable and clear and don't see any errors. Question to think about is how can you extend this structure to implement different types of apps (i.e. a shopping page, weather app, etc.)

PR #549, opened 2018-10-03
React-Todo: Oliver Abreu

AbreuOliver:
@decagondev


AbreuOliver:
thanks, Tom! I'll get right to work 😄

PR #551, opened 2018-10-03
Got it to show - no function - not broken

sweetooth101:
@decagondev

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,16 +1,27 @@
 import React from 'react';
+import TodoForm from './components/TodoComponents/TodoForm'
+import TodoList from './components/TodoComponents/TodoList'
+
 
 class App extends React.Component {
-  // you will need a place to store your state in this component.
-  // design `App` to be the parent component of your application.
-  // this component is going to take care of state, and any change handlers you need to work with your state
-  render() {
-    return (
-      <div>
-        <h2>Welcome to your Todo App!</h2>
-      </div>
-    );
+  constructor() {
+    super();
+    this.state = {
+      todo: [{task:'Sleep',
+              id:564,
+              completed:false,}],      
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -1,16 +1,27 @@
...
+              id:564,
+              completed:false,}],      
+    };
+
   }
+
+
+  render(){
+    return <div>
+            <TodoForm value ={this.state.task} onChange={this.inputHandler}/>
```
decagondev: passing the props correctly including the reference to the handler method

src/components/TodoComponents/TodoForm.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+const Form = props => {
+    return(
+        <div>
+            <input placeholder='..todo'  value={props.todo}></input>
```
decagondev: good use of prop passing for this part

src/components/TodoComponents/TodoForm.js
```diff
@@ -0,0 +1,13 @@
...
+            <button >Add Todo</button>
```
decagondev: remember to add your clicking functionality and remember to have { } curley braces around the clickHandler

src/components/TodoComponents/TodoList.js
```diff
@@ -1,2 +1,15 @@
 // your components will all go in this `component` directory.
 // feel free to change this component.js into TodoList.js
+import React from 'react';
+
+
+const TodoList = props => {
+    return (
+        <div>
+            {props.list.map(item => {
+                return (<div key={item.id}> {item.task} </div>)})}
```
decagondev: good use of item.id as the key

PR #553, opened 2018-10-03
Adam Hinckley

adamhinckley:


src/App.js
```diff
@@ -1,13 +1,48 @@
...
+
+  changeHandler = event => {
+    console.log(event.target.value);
+    this.setState({ newTodo: event.target.value });
+  };
+
+  addNewTodo = event => {
+    event.preventDefault();
+    this.setState({
+      todoList: [...this.state.todoList, { task: this.state.newTodo }],
```
cesarnml: it would make sense to add `id` and `completed` properties in addition to a `task` property to the todo object. more on this on Thursday I'm sure.

src/App.js
```diff
@@ -1,13 +1,48 @@
...
+  };
   // you will need a place to store your state in this component.
   // design `App` to be the parent component of your application.
   // this component is going to take care of state, and any change handlers you need to work with your state
   render() {
     return (
       <div>
-        <h2>Welcome to your Todo App!</h2>
+        <List todoData={this.state.todoList} />
+        <Form addNewTodo={this.addNewTodo} changeHandler={this.changeHandler} newTodo={this.state.newTodo} />
```
cesarnml: good consisting naming

src/components/TodoComponents/TodoForm.js
```diff
@@ -1,5 +1,14 @@
...
-const Form = () => <input type="text" placeholder="Add todo. . ." />;
-
+const Form = props => {
+  return (
+    <form>
+      <input value={props.newTodo} onChange={props.changeHandler} type="text" placeholder="Add todo. . ." />
+      <button type="submit" onClick={props.addNewTodo}>
+        Add Todo
+      </button>
+      <button>Clear Completed</button>
```
cesarnml: clear button is refreshing the page. default behavior should be prevented on thursday

src/App.js
```diff
@@ -1,13 +1,79 @@
...
+    this.setState({
+      todoList: [...this.state.todoList, { task: this.state.newTodo, id: Date.now(), completed: false }],
+      newTodo: ""
+    });
+  };
+  clearCompleted = event => {
+    event.preventDefault();
+    this.setState({
+      todoList: this.state.todoList.filter(task => {
+        return task.completed === true;
```
cesarnml: This isn't working properly. It's clearing all the task that are not completed, instead of clearing the completed task as was requested. Flip the logic.

src/components/TodoComponents/TodoList.js
```diff
@@ -1,2 +1,14 @@
-// your components will all go in this `component` directory.
-// feel free to change this component.js into TodoList.js
+import React from "react";
+import Todo from "./Todo";
+
+const List = props => {
+  return (
+    <div>
+      {props.todoData.map((item, index) => {
+        return <Todo task={item} key={item.task} taskIndex={index} toggleTodo={props.toggleTodo} />;
```
cesarnml: The key attribute must be unique. If I enter the same task twice, the key will not be unique and it will lead to bugs in the UI.

src/App.js
adamhinckley: I fixed that big, just had to change true to false.

src/components/TodoComponents/TodoList.js
adamhinckley: That's something I knew, but didn't implement it there.  I have changed it to id which is unique because it uses Date.now() to create a unique ID for every new take added.  Date.now was already implemented on my addNewTodo item in App.js for that reason.

Thanks for the feedback 👍

PR #554, opened 2018-10-03
Edward Jeong

edjeong99:


src/App.js
```diff
@@ -1,13 +1,58 @@
...
 
 class App extends React.Component {
   // you will need a place to store your state in this component.
   // design `App` to be the parent component of your application.
   // this component is going to take care of state, and any change handlers you need to work with your state
+ constructor(props){
+   super(props);
+   this.state ={items : todoItem};
+  
+   this.addTodo = this.addTodo.bind(this);
```
Velsu: The other way around is would be an arrow function, that way `this` will bind to component implicitly.

PR #559, opened 2018-10-03
Chad Jemmett React ToDo

ceejaay:
@sumi419

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -17,17 +17,19 @@ class App extends React.Component {
          id: Date.now(),
          complete: false}
       ],
-      newItem: ''
+      newItem: '',
+      newThing: 1 
+
       }
 
   }
-  // you will need a place to store your state in this component.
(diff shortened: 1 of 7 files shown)


src/App.js
```diff
@@ -1,15 +1,86 @@
...
+
+      }
+  }
+
+  crossOut = (index) => {
+    this.setState({list: this.state.list.map((item, idx) => {
+      if(index === idx) {
+        return {
+          ...item,
+          complete: item.complete === false ? true : false,
```
Velsu: Another possibility here would be to use a toggler like `complete: !item.complete`

src/components/TodoComponents/Todo.js
```diff
@@ -0,0 +1,11 @@
+
+import React from 'react';
+const Todo = props => {
+  return (
+      <div>
```
Velsu: If this component is not doing anything it can be safely removed for the time beign.

PR #560, opened 2018-10-03
Zachary Arney

zacharyarney:


src/App.js
```diff
@@ -1,16 +1,93 @@
...
+        return todo;
+      }),
+    });
+  };
+
+  clearCompleted = (event) => {
+    event.preventDefault();
+    this.setState({
+      todos: this.state.todos.filter((todo) => {
+        return todo.completed === false;
```
luiscmartinez: a clean syntax sugary way would be just to implicitly return 
this.state.todos.filter((todo) => todo.completed === false)

src/components/TodoComponents/TodoForm.js
```diff
@@ -0,0 +1,28 @@
...
+    <form className="todo-form" onSubmit={props.addTodoItem}>
+      <div className="submit-form">
+        <input
+          type="text"
+          name="newTodoItem"
+          placeholder="What todo?"
+          value={props.text}
+          onChange={props.changeHandler}
+        />
+        <button type="submit" onClick={props.addTodoItem}>
```
luiscmartinez: this button wouldn't need an onClick because it is nested inside the Form tag therefore the button will automatically be bind to the onSubmit attribute on the Form element when being click on this button or hitting the enter key after typing on the input . All because they are nested inside the form tag

PR #564, opened 2018-10-03
Ryan Clausen

theryanclausen:


CookieMonsta89:
Awesome work with you todo list Ryan.  I like that it has the red font and line through when checked.  You did a great job passing props and state.  You also did a pretty good job of keeping components smaller but you could have made them even smaller by using the todo.js file.  As applications get larger, you will want the components to be broken down as much as possible.  Great work man

PR #570, opened 2018-10-03
CJ Wright

cwright0428:
proj start and PR

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,14 +1,71 @@
-import React from 'react';
+import React from "react";
+import TodoList from "./components/TodoComponents/TodoList.js";
+import TodoForm from "./components/TodoComponents/TodoForm.js";
+import styled from "styled-components";
+import { width } from 'window-size';
+
+
+const wrapper = styled.div`
+  height:500px;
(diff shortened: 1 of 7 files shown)


src/App.js
```diff
@@ -1,14 +1,71 @@
-import React from 'react';
+import React from "react";
+import TodoList from "./components/TodoComponents/TodoList.js";
+import TodoForm from "./components/TodoComponents/TodoForm.js";
+import styled from "styled-components";
+import { width } from 'window-size';
+
+
+const wrapper = styled.div`
```
Velsu: This needs to be uppercased `Wrapper` to work properly.

src/App.js
```diff
@@ -1,14 +1,71 @@
...
+      inputVal: ""
+    }));
+  };
+
   render() {
     return (
-      <div>
-        <h2>Welcome to your Todo App!</h2>
-      </div>
+      <wrapper>
```
Velsu: The styled div's name needs to be uppercased to work like mentioned above, You want to do `<Wrapper></Wrapper>`

PR #571, opened 2018-10-03
Ryan Walker - React-Todo

rytwalker:


src/App.js
```diff
@@ -1,13 +1,126 @@
...
+
+  handleNewToDo = e => {
+    e.preventDefault();
+    // !this.state.input is equivalent to this.state.input === '' (falsey)
+    if (!this.state.input) {
+      this.setState({ invalidInput: true });
+    } else {
+      const task = {
+        // slice creates a copy of this.state.input
+        task: this.state.input.slice(),
```
kamry-bowman: don't need to create a copy of this.state.input, it is a string, and they are immutable, so they are always copied 

https://codeburst.io/explaining-value-vs-reference-in-javascript-647a975e12a0

src/App.js
```diff
@@ -1,13 +1,126 @@
...
+        completed: false
+      };
+
+      // copy todos and add task to todos
+      const todos = [...this.state.todos, task];
+
+      // add task to todo list
+      // todos.push(task);
+
+      this.setState({ todos: todos, input: '' });
```
kamry-bowman: you can do this.setState({ todos, input: '' }

https://ariya.io/2013/02/es6-and-object-literal-property-value-shorthand

src/App.js
```diff
@@ -1,13 +1,126 @@
...
+          // handle empty string
+          this.setState({ [key]: value });
+        }
+      }
+    }
+  }
+
+  // React.Component method
+  componentDidMount() {
+    this.hydrateStateWithLocalStorage();
```
kamry-bowman: very cool!

src/App.js
```diff
@@ -1,13 +1,126 @@
...
+      <div className="App">
+        <TodoHeader />
+        <TodoForm
+          handleNewToDo={this.handleNewToDo}
+          handleInputChange={this.handleInputChange}
+          handleCompletedTasks={this.handleCompletedTasks}
+          input={this.state.input}
+        />
+        <TodoMessage
+          inlineStyle={this.state.invalidInput ? this.state.showMessage : null}
```
kamry-bowman: Love it!

kamry-bowman: However, the specific pattern of having this.state.showMessage on state, not sure the reason? It never gets updated, and I'm not sure if you would want to update it. What I would do is just have a CSS class called `.hidden`  that sets it to `display: none` and have a prop on TodoMessage that checks for whether invalidInput is on state, and use that prop to determine whether to include 'hidden' as className inside of TodoMessage by using a short circuit operator. Like this:

`<TodoMessage visible={Boolean(this.state.invalidInput)} />`

and then inside do:

```
const TodoMessage = props => {
  return (
    <div className="TodoMessage" className={props.visible && 'hidden'} >
      <i className="fas fa-exclamation-triangle" />
      Enter a task, s'il vous plaît!
    </div>
  );
};
```

src/App.js
```diff
@@ -1,13 +1,126 @@
...
+        />
+        <TodoList
+          todos={this.state.todos}
+          toggleCompleteTask={this.toggleCompleteTask}
+          style={this.state.todoStyle}
+          inlineStyle={this.state.toDoCompleteStyle}
```
kamry-bowman: I wouldn't put this on state... it would not be something that was fluctuating. If you want to have an inlineStyle prop on your TodoList to make the representation flexible, just pass that as a prop with a string, or just have some sort of variable or import that you are keeping that theming info with.

But in my opinion CSS already handles this, you use the React and classnames to handle the "business logic" and you use the CSS to define what that actually looks like on the screen. So that's my preferred approach.

Since you're exploring patterns like this, I would say you will likely find styled-components very compelling when we learn about those, and other CSS-in-JS options.

I would say in any case don't put it on state though, state is for things that dynamically update. What dynamically updates is whether something is in an error state, or whether it's completed, not *how* that is visually represented.

PR #572, opened 2018-10-03
Justin Lowry: React-Todo

dividedsky:


src/components/TodoComponents/Todo.js
```diff
@@ -1,18 +1,12 @@
...
-//   }
-
-
-      // {/* className={props.todo.completed} */}
 const Todo = props => {
   return (
-    <div className="todo"
-    onClick={props.toggleCompleted}
-    id={props.todo.id}
+    <div
```
jeffyjkang: like this double ternary action

PR #573, opened 2018-10-03
Angelo Deleon - React-Todo

RaDeleon:
@kamry-bowman

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,39 +1,53 @@
 import React from 'react';
 import TodoList from './components/TodoComponents/TodoList';
 import TodoForm from './components/TodoComponents/TodoForm';
-import Todo from './components/TodoComponents/Todo';
+import './App.css';
 
-
-
-let todoItem = [{task : '#1', id : Date.now(),  completed : false}];
+let todoItem = [{task : '1st', id : Date.now(),  completed : false}];
 
 class App extends React.Component {
-  constructor(props){
-    super(props);
-    this.state = { items: todoItem};
-    this.addTodo = this.addTodo.bind(this);
+
+ constructor(props){
+   super(props);
+   this.state ={items : todoItem};
+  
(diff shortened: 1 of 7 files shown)


src/App.js
```diff
@@ -1,14 +1,39 @@
...
-  // design `App` to be the parent component of your application.
-  // this component is going to take care of state, and any change handlers you need to work with your state
+  constructor(props){
+    super(props);
+    this.state = { items: todoItem};
+    this.addTodo = this.addTodo.bind(this);
+  }
+
+  addTodo(item) {
+    let  newItem = {task: item, id: Date.now(), completed: false };
```
kamry-bowman: this isn't quite right, you need to get the value for task from `this.state`, and to do that, you need to get the input form to be updating state with the string you're entering

src/App.js
```diff
@@ -1,14 +1,39 @@
...
+ 
   render() {
+    console.log(`App render ${this.state.items}`);
+    console.log(this.state.items);
     return (
       <div>
-        <h2>Welcome to your Todo App!</h2>
+
+        <h2>Welcome to the Best Todo App Ever!</h2>
+        <TodoList key={this.state.id}  items={this.state.items} />
```
kamry-bowman: I don't think you need a `key` value here

src/components/TodoComponents/TodoForm.js
```diff
@@ -0,0 +1,34 @@
...
+
+class TodoForm extends React.Component {
+    constructor(props){
+        super(props);
+        this.onSubmit = this.onSubmit.bind(this);
+    }
+
+    onSubmit(e){
+        e.preventDefault();
+        let newItem = this.refs.item.value;
```
kamry-bowman: using refs isn't the desired solution here, you want to pass a function down from App.js on props and use that in this on `<form>` for `onSubmit`

PR #575, opened 2018-10-03
Hargobind Atwal

ethiapath:


CookieMonsta89:
Great job getting your todo list to work for day 1 mvp.  I'm not sure if you are doing it on purpose or not but there is a true or false displayed after each todo.  This is due to you setting the ternary operator to strings but I suppose you may know that.  You can create a function that changes text decoration and use it instead of the true/false text or you can create the style right in the ternary operator.

PR #582, opened 2018-10-03
Conner H [React-ToDo]

TRIF3X:


CookieMonsta89:
Good job Conner getting your day 1 mvp finished.  So the reason you have the array of objects in the state in app.js is so that they display to screen and you can 'push' your new tasks to the list.  You have to objects but the tasks are empty strings.  No big deal, just wanted to point out what that was doing.  I like that you used all three todo files to keep components smaller.  Good work.

PR #583, opened 2018-10-03
React-Todo- Brian Gregorius

brianmgre:
@CookieMonsta89


CookieMonsta89:
Great job getting your todo list to work for day 1 mvp.  Great job passing props and state through components.  On the todo.js file, you will have an if statement that will check if an item has been clicked and if so, it'll apply style with line through.  Hope this helps with day 2.  Keep it up.

PR #585, opened 2018-10-03
David Soto React-ToDo

David247365:
@CookieMonsta89

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -1,0 +1,8 @@
+.to-do-list {
+  height: 80vh;
+  width: 100%;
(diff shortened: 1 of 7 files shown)


CookieMonsta89:
Great job getting the todo list day mvp done.  I would recommend trying to split things up into different files.  Try to utilize todo.js, todoform.js and todolist.js.  On the form just pass props through to it from app.js so that it knows what the data is and on todolist, you'll map over the data to display it on screen.  The todo.js you will use when working with the day 2 mvp.  Hope this helps

PR #592, opened 2018-10-03
Andre Myrick - React-Todo

AM-Myrick:


src/components/TodoComponents/Todo.js
```diff
@@ -0,0 +1,9 @@
+import React from "react";
```
CookieMonsta89: You'll be using this file to finish day 2 mvp.  You will pass a function through here that will check to see if an item was completed, if so, it'll get a style of line through, if not, it'll be null

PR #593, opened 2018-10-03
 Troy Amlee [mvp] completed font-style on-click, and remove

troyAmlee:


src/App.js
```diff
@@ -1,13 +1,129 @@
...
+        id: 1598817077186,
+        completed: "false"
+      },
+      ],
+      newTask: '',
+      newId: Math.floor((Math.random() * 1000000000000) + 1000000000000),
+      newCompleted: "false",
+      newDeleted: "true",
+      textDecoration: 'line-through',
+    }
```
cesarnml: only dynamic variables should be on `state` newCompleted, newDeleted, textDecoration won't change so they should be state variables

src/App.js
```diff
@@ -1,13 +1,129 @@
...
+    // this.setState({ completed: !this.state.completed });
+
+    this.setState({ 
+      toDoList: this.state.toDoList.map((item, idx)=>{
+        if (id !== idx){
+          return item;
+        } else {
+          return {
+            ...item,
+            completed: item.completed === "false" ? "true" : "false"
```
cesarnml: `completed: !item.completed` is a cleaner approach

src/App.js
```diff
@@ -1,13 +1,129 @@
...
+          };
+        }
+      })
+     });
+  };
+
+  clearTask = event => {
+    console.log(event.target.name);
+    let targeted = document.getElementsByClassName('toggle-decoration');
```
cesarnml: review use cases for DOM manipulation when using React in the docs

src/App.js
```diff
@@ -1,13 +1,129 @@
...
+  // event.preventDefault();
+  // this.setState({
+   
+  // })
+  };
+
+  filterCharacter = () => {
+   
+        this.setState({
+          toDoList: this.state.toDoList.filter((item) => {
```
cesarnml: `item => !item.completed` is a cleaner implementation of .filter

PR #604, opened 2018-10-03
Brandon Grych React-Todo

brandongrych:


src/App.js
```diff
@@ -1,16 +1,115 @@
...
+
+  changeHandler = event => {
+    const value = event.target.value;
+    this.setState({
+      input: value,
+    })
+  }
+  clickHandler = (event, item) => {
+    event.preventDefault();
+    if (event.target.classList.contains('strikethrough')) {
```
Velsu: I have a feeling way to toggle todo's as complete or not could be done a little better and more efficient, especially with that loop. There could be some toggle method that will change the current state for single todo without need of loop.

PR #606, opened 2018-10-03
Brandon Vilaychith React TODO

BrandonVilaychith:


src/components/TodoComponents/Todo.css
```diff
@@ -0,0 +1,110 @@
+* {
+    box-sizing: border-box;
+    margin: 0;
+    padding: 0;
+}
+body {
+    background-image: url('https://images.unsplash.com/photo-1538617648085-8cd6adae0835?ixlib=rb-0.3.5&ixid=eyJhcHBfaWQiOjEyMDd9&s=d079ea4a9ecc9bd19ecced59bc0ce005&auto=format&fit=crop&w=1534&q=80');
+    background-size: cover;
+    background-position: center;
```
cesarnml: `background-repeat: no-repeat` is essentially, so that your website doesn't look like it's from 1982

PR #609, opened 2018-10-03
Jacob A Brennan

jacobabrennan:


src/ProgressTracker.js
```diff
@@ -0,0 +1,104 @@
+import React from 'react';
+import ToDoList from './components/ToDoComponents/ToDoList.js';
```
Velsu: There are some issues with paths for importing components and css files in multiple places.

PR #618, opened 2018-10-03
Carlo Clamucha - React To Do list

CarloC24:


KevinBrack:
Nice work. Know we talked about it but I wanted to remind you that if you use `<form onSubmit={() => cb()} />` you can get the same result of triggering upon [enter] and [submit button]

PR #619, opened 2018-10-03
React-Todo Edd Burke

bummings:


KevinBrack:
Nice work. I do not see any select, or delete functionality at this point. Also would be nice to have the input field reset after submission. Need any help let me know.

PR #630, opened 2018-10-04
todo

codypeak:


KevinBrack:
Nice work. We need some more explanation in commit messages. These are not sufficient. I do not see any select or delete functionality at this point.

PR #635, opened 2018-10-10
Jake Thomas

jmaxt12:


src/components/TodoComponents/Todo.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+// import ToDo from './'
+
+const ToDo = (props) => {
+    return (
+        <div>
+            {props.toDoTasks.map((item, index) => {
+                return (
+                    <h1
```
Velsu: The `h1` tag is not closed which will cause errors.

PR #639, opened 2018-11-07
Sean Pheneger: WIP

wcolts2000:
initial commit, initializing pr sequence

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,18 +1,9 @@
 import React from 'react';
+import './components/TodoComponents/Todo.css'
 import TodoList from './components/TodoComponents/TodoList';
 import TodoForm from './components/TodoComponents/TodoForm';
 
 const todoListSeed = [
-  // {
-  //   task: 'Organize Garage',
-  //   id: 1528817077286,
-  //   completed: false
-  // },
-  // {
-  //   task: 'Bake Cookies',
-  //   id: 1528817084358,
-  //   completed: false
-  // }
   {
     task: 'Learn setState()',
     id: Date.now(),
@@ -58,16 +49,6 @@ class App extends React.Component {
         ],
           inputText: '',
       })
-
-
-
-    // this.setState({
-    //   todoList: [
-    //     ...this.state.todoList,
-    //     { task: this.state.inputText}
-    //   ],
-    //   inputText: ''
-    // });
   };
 
   handleChange = event => {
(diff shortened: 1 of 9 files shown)


src/App.js
```diff
@@ -1,13 +1,89 @@
 import React from 'react';
```
szincone: If you write `import React from 'react';` as `import React, { Component } from 'react';` line 19 could be written as `class App extends Component {` since you imported a named import you wouldn't have to use that dot.syntax. Either way is fine, just showing alternate ways.

src/App.js
```diff
@@ -1,13 +1,89 @@
...
+    this.state = {
+      todoList: todoListSeed,
+      inputText: '',
+      id: Date.now(),
+      completed: false
+    }
+  }
+
+  componentDidMount() {
+    const storedTodo = JSON.parse(localStorage.getItem('todo'));
```
szincone: You are messing w/ my local storage, I don't know if it's what you're trying to save to local storage, but it is definitely saving stuff to my local storage.

src/App.js
```diff
@@ -1,13 +1,89 @@
...
+      })
+  };
+
+  handleChange = event => {
+    this.setState({
+      [event.target.name]: event.target.value
+    });
+  };
+
+  toggleCompleted = (event, todo) => {
```
szincone: Good job on getting your line-through styling working.

src/App.js
```diff
@@ -1,13 +1,89 @@
...
+    // todo.setState({completed: false })
+    this.setState(todo => {
+      return { completed: (todo.completed: true)}
+    })
+    console.log(todo)
+  }
+
   render() {
     return (
       <div>
```
szincone: Great place to use a `<Fragment>` since you're only using this div as a container for your components and not using it for any styling.

src/components/TodoComponents/Todo.js
```diff
@@ -0,0 +1,20 @@
...
+
+const Todo = props => {
+
+  
+  return (
+    <div>
+      
+      <p onClick={(e) => {
+        // {console.log(props)}
+        props.toggleCompleted(e, props)
```
szincone: Good job on implementing both params passed through your toggleCompleted handler.

src/components/TodoComponents/Todo.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+import './Todo.css'
+
+const Todo = props => {
+
+  
+  return (
+    <div>
```
szincone: Another good spot for a `<Fragment>`

PR #650, opened 2018-11-07
Nick Stricker

NickStrick:


src/App.js
```diff
@@ -1,16 +1,110 @@
...
+    })
+  }
+
+  clearCompleted = ev =>{
+    let indexArr = [];
+    let tempList = this.state.itemList.slice();
+    for(let i = 0; i<tempList.length;){
+      if(tempList[i].completed){
+         
+          indexArr.push(i);
```
johnoro: Some of your indentation and spacing is off; [prettier](https://prettier.io) can format your code for you.

src/App.js
```diff
@@ -1,16 +1,110 @@
...
+    super();
+    this.state = {
+      itemList: listData,
+      inputText: '',
+    }
+  }
+
+  handleChange = ev =>{
+    this.setState({
+      [ev.target.name]: ev.target.value
```
johnoro: Consider [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring) here to make this easier to write and read:
```javascript
const { name, value } = ev.target;
```

src/components/TodoComponents/TodoList.js
```diff
@@ -6,7 +6,7 @@ import './Todo.css';
 
 const TodoList = props => {
     return (
-        <div>
+        <div className={'todoList'}>
```
johnoro: Why use braces here to evaluate the string? You can use just the string just fine:
```javascript
<div className='todoList'>
```
I'd also recommend keeping CSS to be kebab-cased over camelCased (so `todo-list` over `todoList` for the CSS class here).

PR #651, opened 2018-11-07
First Submission - Day One MVP Completed

gittc100:


src/components/TodoComponents/TodoForm.js
```diff
@@ -0,0 +1,21 @@
...
+          type="text"
+          name="inputText"
+          placeholder="Add New Todo"
+          value={props.inputText}
+          onChange={props.handleChange}
+        />
+        <button type="submit">Add Todo</button>
+        <button>Clear Completed</button>
+      </form>
+    </React.Fragment>
```
johnoro: Why use a fragment here? Your form is the parent element and has no siblings; the fragment adds nothing dom/styling-wise, so it's unnecessary here.

src/components/TodoComponents/Todo.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import './Todo.css';
+
+const Todo = props => {
+  return (
+    <p className="list-text">{props.toDoItem.task}</p>
```
johnoro: Consider using [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring) here:
```javascript
const Todo = ({ toDoItem }) => {
  const { task } = toDoItem;
  return <p className="list-text">{task}</p>;
};
```

src/App.js
```diff
@@ -1,13 +1,58 @@
...
+    super();
+    this.state = {
+      toDoItems: ToDo,
+      inputText: ""
+    };
+  }
+
+  handleChange = event => {
+    this.setState({
+      [event.target.name]: event.target.value
```
johnoro: Consider object destructuring for this method:
```javascript
handleChange = ({ target }) => {
  const { name, value } = target;
};
```

src/components/TodoComponents/Todo.js
```diff
@@ -3,8 +3,9 @@ import './Todo.css';
 
 const Todo = props => {
   return (
-    <p className="list-text">{props.toDoItem.task}</p>
+    <p className={`list-text ${props.toDoItem.completed}`} onClick={() => props.changeStyle(props.toDoItem.id)} >{props.toDoItem.task}</p>
```
johnoro: This looks like it ought to be broken up for readability:
```javascript
<p
  className={`list-text ${props.toDoItem.completed}`}
  onClick={() => props.changeStyle(props.toDoItem.id)} 
>
  {props.toDoItem.task}
</p>
```
I'd also recommend some [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring) here:
```javascript
const Todo = ({ changeStyle, todoItem }) => {
  const { completed, id, task } = todoItem;
  return (
    <p
      className={`list-text ${completed}`}
      onClick={() => changeStyle(id)} 
    >
      {task}
    </p>
  );
};
```

PR #654, opened 2018-11-07
Zayna Crick

ZaynaCrick:


src/App.js
```diff
@@ -1,16 +1,86 @@
...
+    this.state = {
+      list: listData,
+      inputText: '',
+    };
+  }
+  handleChange = event => {
+    this.setState ({
+      [event.target.name]: event.target.value
+    });
+    console.log(event.target.name);
```
trevorfehrman: I love to see consolelogs, gj

src/App.js
```diff
@@ -1,16 +1,86 @@
...
+  };
+  
+  addItem = ev => {
+    ev.preventDefault();
+    this.setState ({
+      list: [
+        ...this.state.list,
+        {task: this.state.inputText, id: Date.now(),completed: 'false' }
```
trevorfehrman: The only danger in using Date.now for ids is if you have an application that makes many items at once you can end up with non-unique ids.  This works perfectly for this application though.

src/components/TodoComponents/Todo.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import TodoList from './TodoList';
+
+const Todo = props => {
+    return (
+    <div>
+        {props.list.map(listsInMap => (
```
trevorfehrman: i like your semantic naming of your cb parameter here gj thinking about what the data your passing around is

PR #657, opened 2018-11-07
Nick O'Ferrall

nickoferrall:


src/components/TodoComponents/TodoForm.js
```diff
@@ -0,0 +1,64 @@
...
+//     return (
+//         <div>
+//             <form onSubmit={props.addToDo} placeholder="Enter stuff here"></form>
+//             <button>Add Todo</button>
+//             <button>Clear Completed</button>
+//         </div>
+//     )
+// }
+
+// const TodoForm = props => {
```
szincone: I know leaving comments in isn't the best thing in a professional setting but I think it's really good when you're learning so you can always go back and reference your code.

src/components/TodoComponents/TodoForm.js
```diff
@@ -0,0 +1,64 @@
...
+//           />
+//           <button type="submit">Add Character</button>
+//         </form>
+//       </div>
+//     );
+//   };
+
+const CharacterForm = props => {
+    return (
+      <div>
```
szincone: This would be a great place to use a `<Fragment>` since all your using this div for is a wrapper for containing your form and not using it for styling or anything.

PR #659, opened 2018-11-07
Grace Kang

bananaboots:


src/App.js
```diff
@@ -1,13 +1,83 @@
...
+  }
+
+  changeHandler = (event) => {
+    this.setState({
+      inputText: event.target.value
+    })
+  }
+
+  completedToggle = (event) => {
+    event.target.classList.toggle('completed');
```
johnoro: Consider object destructuring for this function:
```javascript
completedToggle = ({ target }) => {
  const { classList, key } = target;
  classList.toggle('completed');
  // ...
};
```

src/components/TodoComponents/TodoList.js
```diff
@@ -1,2 +1,13 @@
-// your components will all go in this `component` directory.
-// feel free to change this component.js into TodoList.js
+import React from 'react';
+import Todo from './Todo';
+
+function TodoList(props) {
+    return (
+        <div className="list">
+            <h1>To Do:</h1>
+            {props.todoArray.map(x => <Todo completedToggle={props.completedToggle} task={x.task} key={x.id} id={x.id} />)}
```
johnoro: This would do better split up for readability as well:
```javascript
{props.todoArray.map(x => <Todo 
  completedToggle={props.completedToggle} 
  task={x.task} 
  key={x.id} 
  id={x.id} 
/>)}
```

src/App.js
```diff
@@ -1,3 +1,11 @@
+//      _                  _     
+//     / \   _ __  _ __   (_)___ 
+//    / _ \ | '_ \| '_ \  | / __|
+//   / ___ \| |_) | |_) | | \__ \
+//  /_/   \_\ .__/| .__(_)/ |___/
+//          |_|   |_|   |__/
```
johnoro: I like this, but you probably wouldn't want this in production.

src/App.js
```diff
@@ -13,25 +21,29 @@ class App extends React.Component {
...
+          {
+            task: this.state.inputText, 
+            id: Date.now(), 
+            completed: false
+          },
+          ...JSON.parse(window.localStorage.getItem('tasks'))
+        ]
+      }, () => { window.localStorage.setItem('tasks', JSON.stringify(this.state.tasks)); return this.forceUpdate() });
+      this.setState({
+        inputText: ''
```
johnoro: Why have this in a separate `setState`?

src/App.js
```diff
@@ -13,25 +21,29 @@ class App extends React.Component {
...
   }
 
   clearTasks = (event) => {
+    event.preventDefault();
+    console.log('buttons working dude');
     this.setState({
-      tasks: this.state.tasks.filter(x => Object.values(x).includes(false))
-    })
+      tasks: JSON.parse(window.localStorage.getItem('tasks')).filter(x => Object.values(x).includes(false))
+    }, () => { window.localStorage.setItem('tasks', JSON.stringify(this.state.tasks)); return this.forceUpdate() })
```
johnoro: It looks like you have this in multiple places; why not make a method for it? For example:
```javascript
clearTasks = event => {
  // ...
  this.setState({
    tasks: JSON.parse(... , 
    this.setTasks();
  // ...
};

setTasks = () => {
  window.localStorage
    .setItem('tasks', JSON.stringify(this.state.tasks));
  return this.forceUpdate();
};
```

src/components/TodoComponents/Todo.js
```diff
@@ -1,8 +1,15 @@
...
+//    | | (_) | |_| | (_) | 
+//    |_|\___/|____/ \___(_)
+                         
+
 import React from 'react';
 import './Todo.css';
 
 function Todo(props) {
-    return <p onClick={props.completedToggle}>{props.task}</p>
+    return <p className={`card ${props.completed}`} onMouseOver={props.pointer} onClick={() => props.completedToggle(props.id)} >{props.task}</p>
```
johnoro: You should definitely break this up for readability (which [prettier](https://prettier.io) would do for you):
```javascript
return (
  <p
    className={`card ${props.completed}`}
    onMouseOver={props.pointer} 
    onClick={() => props.completedToggle(props.id)}
  >
    {props.task}
  </p>
);
```
and here's a version assuming you've used [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring):
```javascript
return (
  <p
    className={`card ${completed}`}
    onMouseOver={pointer} 
    onClick={() => completedToggle(id)}
  >
    {task}
  </p>
);
```

src/components/TodoComponents/TodoList.js
```diff
@@ -1,11 +1,18 @@
...
 import React from 'react';
 import Todo from './Todo';
 
 function TodoList(props) {
     return (
         <div className="list">
-            <h1>To Do:</h1>
-            {props.todoArray.map(x => <Todo completedToggle={props.completedToggle} task={x.task} key={x.id} id={x.id} />)}
+            <div className="h1"><h1>To Do:</h1></div>
+            {props.todoArray.map(x => <Todo completedToggle={props.completedToggle} task={x.task} key={x.id} id={x.id} pointer={props.pointer} completed={x.completed}/>)}
```
johnoro: Break these two lines up for readability!

PR #662, opened 2018-11-07
Jonathan Heinz

jheinz1999:


src/App.js
```diff
@@ -1,13 +1,127 @@
...
+
+      todoItems: todoList,
+      formInput: '',
+      searchInput: '',
+      filteredItems: todoList
+
+    }
+
+    this.btnClick = this.btnClick.bind(this);
+    this.formEntry = this.formEntry.bind(this);
```
johnoro: You could alternatively use arrow functions for your methods. For example:
```javascript
btnClick = e => {
  e.preventDefault();
  // ...
};
```

src/App.js
```diff
@@ -1,13 +1,127 @@
...
+
+      localStorage.clear();
+
+    }
+
+  }
+
+  formEntry(e) {
+
+    const name = e.target.name;
```
johnoro: Why not use [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring) here? For example:
```javascript
const { name, value } = e.target;

if (name == "searchInput") {
// ...
```

src/App.js
```diff
@@ -1,13 +1,127 @@
...
+
+  }
+
   render() {
     return (
-      <div>
-        <h2>Welcome to your Todo App!</h2>
+      <div className='todo-container'>
+        <h1>React Todo App</h1>
+        <TodoForm entryHandler={this.formEntry} btnHandler={this.btnClick} formInput={this.state.formInput} searchInput={this.state.searchInput}/>
```
johnoro: It'd be a good idea to break this up for readability since you have a lot of props. For example:
```javascript
<TodoForm
  entryHandler={this.formEntry} 
  btnHandler={this.btnClick} 
  formInput={this.state.formInput} 
  searchInput={this.state.searchInput}
/>
```

src/components/TodoComponents/Todo.js
```diff
@@ -2,11 +2,11 @@ import React from 'react';
 
 import './Todo.scss';
 
-const Todo = ({item}) => {
+const Todo = ({item, clickHandler}) => {
 
   return (
 
-    <li>{item.task}</li>
+    <li onClick={() => clickHandler(item.id)} className={item.completed ? 'completed' : null}><input type='checkbox' onChange={() => {}} checked={item.completed} /> {item.task} <p></p></li>
```
johnoro: You should certainly move the attributes here onto separate lines for readability:
```javascript
<li 
  onClick={() => clickHandler(item.id)} 
  className={item.completed ? 'completed' : null}
>
  <input 
    type='checkbox' 
    onChange={() => {}} checked={item.completed} 
  /> 
  {item.task} 
  <p></p>
</li>
```
I had no idea what all was going on here until I broke up the tags and the attributes.

johnoro: You could also use object destructuring even moreso via:
```javascript
const { completed, id, task } = item;
```
and I assume you can imagine that the code that follows would be a bit shorter.

PR #666, opened 2018-11-08
Chaya Otikor: Stretch

cotikor:


src/App.js
```diff
@@ -1,14 +1,57 @@
 import React from 'react';
```
szincone: if you write `import React from 'react';` as `import React, { Component } from 'react';`  line 19 could be written as `class App extends Component {` since you imported a named import you wouldn't have to use that dot.syntax. Either way is fine, just showing alternate ways.

src/App.js
```diff
@@ -1,14 +1,57 @@
...
+  constructor(props){
+    super(props);
+    this.state = {
+      list: listItems,
+      inputText: '',   
+    }
+  }
+  handleChange = event => {
+    this.setState({
+      [event.target.name]: event.target.value
```
szincone: Good job getting your controlled inputs working.

PR #667, opened 2018-11-08
Ben Tsao

cbtsao47:


src/App.js
```diff
@@ -1,13 +1,56 @@
-import React from 'react';
+import React from "react";
```
szincone: If you write import React from 'react'; as import React, { Component } from 'react'; line 18 could be written as class App extends Component { since you imported a named import you wouldn't have to use that dot.syntax. Either way is fine, just showing alternate ways.

src/App.js
```diff
@@ -1,13 +1,56 @@
...
+    this.setState(
+          {
+           toDos:[...this.state.toDos, {task: this.state.inputText}],
+           inputText :''
+          }
+    );
+  }
   render() {
     return (
       <div>
```
szincone: Great candidate for a `<Fragment>` here since it's just a wrapping div and you're not using it for anything other than wrapping.

src/App.js
```diff
@@ -1,13 +1,56 @@
...
-        <h2>Welcome to your Todo App!</h2>
+        <TodoList toDoList={this.state.toDos}/>
+        <TodoForm 
+        inputText={this.state.inputText} 
+        handler = {this.handler}
```
szincone: spacing inconsistency on this line and next line.

PR #669, opened 2018-11-08
Max David Metelus :MVP Complete

Mdmetelus:
Work in progress

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -24,14 +24,15 @@ class App extends React.Component {
       this.state = {
         allStuffToDo: stuffToDo,
         inputText: '',
-        inputTextThree: '',
-        inputTextTwo: '',
+        
+        // inputTextThree: '',
+        // inputTextTwo: '',
       };
     }
     
     handlesChange = event => {
       this.setState({
-        [event.target.task]: event.target.value
+        inputText: event.target.value
       });
       
       console.log(event.target.task);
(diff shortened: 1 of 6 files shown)


src/App.js
```diff
@@ -1,16 +1,93 @@
...
+    
+    handleChange = event => {
+      this.setState({
+        [event.target.name]: event.target.value
+      });
+      
+      // console.log(event.target.name);
+    };
+    
+    addTask = event => {
```
trevorfehrman: conventionally methods like these begin with the word handle, this is purely opinion i just thought you might want to be aware of that

src/components/TodoComponents/Todo.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+
+const Todo = props => { 
+    return (
+        <h2
+        onClick={() => props.changeFinished(props.todoOnProp.id)}
+        className={`eachListed ${props.todoOnProp.completed}`}>
```
trevorfehrman: I like to see how semantic your names are, you'll slip into your own convention as time goes on but it's awesome to see that you're actually thinking about what all these values are

src/components/TodoComponents/TodoList.js
```diff
@@ -1,2 +1,20 @@
-// your components will all go in this `component` directory.
-// feel free to change this component.js into TodoList.js
+import React from 'react';
+import Todo from './Todo';
+
+const TodoList = props => {
+    return(
+        <div>
+            {props.tasks.map(eachOneInMap => (
```
trevorfehrman: You could use filter here to implement a basic search.  You'll get a bunch of chances to do that in the coming weeks.

PR #671, opened 2018-11-08
Daniel Mendez Class-Components I & II

dan2014:


src/App.js
```diff
@@ -1,16 +1,98 @@
 import React from 'react';
+import TodoList from './components/TodoComponents/TodoList';
+import TodoForm from './components/TodoComponents/TodoForm';
+
+
+if(localStorage.getItem('myData')===null){
+  var TodoData =[];
+} else {
+  TodoData = JSON.parse(localStorage.getItem('myData'));
```
trevorfehrman: You could try this in a componendidmount

src/App.js
```diff
@@ -1,16 +1,98 @@
...
+      todos: TodoData,
+      inputText:""
+    };
+  }
+
+  addTodo = event => {
+    event.preventDefault();
+    this.setState({
+      todos:[
+        ...this.state.todos,{task:this.state.inputText,id:Date.now(),
```
trevorfehrman: This is good.  Looking forward you need to be careful about using Dante.now() as the id.  If your application creates a bunch of things at once you could end up with non-unique ids.  In this case it works perfectly but in other cases you just want to be aware of its limitations.

src/components/TodoComponents/Todo.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+const Todo = props => {
+    return(
+        <React.Fragment>
+            <p className="para" onClick={props.strike}>{props.task}</p>
```
trevorfehrman: Clean

src/components/TodoComponents/TodoList.js
```diff
@@ -1,2 +1,19 @@
-// your components will all go in this `component` directory.
-// feel free to change this component.js into TodoList.js
+import React from 'react';
+import Todo from '../TodoComponents/Todo'
+
+const TodoList = props => {
+    return(
+        <div>
+        {
+            props.todo.map(ele => {
```
trevorfehrman: You could use filter here to implement search.  You'll get opportunities to do that next week.

PR #672, opened 2018-11-08
Brandon Campbell

BrandCam:
was a little under the weather ill finish tn

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import TodoList from './components/TodoComponents/TodoList';
 import TodoForm from './components/TodoComponents/TodoForm';
+import './App.css';
 //data
 const TodoData = [
   {
--- a/src/App.js
+++ b/src/App.js
@@ -5,15 +5,22 @@ import TodoForm from './components/TodoComponents/TodoForm';
 const TodoData = [
   {
     task: 'Organize Garage',
-    id: 1528817077286,
+    id: 1,
     completed: false
   },
   {
     task: 'Bake Cookies',
-    id: 1528817084358,
+    id: 2,
     completed: false
   }
 ];
+
+let nextId = 2;
+
+const getNewId = () => {
+  nextId += 1;
+  return nextId;
+};
 
 class App extends React.Component {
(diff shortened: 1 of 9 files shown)


src/App.js
```diff
@@ -1,13 +1,62 @@
...
+    todos: TodoData,
+    inputText: '' 
+
+
+  }
+}
+
+ handleChange = e => {
+  this.setState({
+    [e.target.name]: e.target.value
```
johnoro: You could use [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring) here for brevity:
```javascript
const { name, value } = e.target;
```

src/components/TodoComponents/TodoList.js
```diff
@@ -1,2 +1,21 @@
...
+
+function TodoList(props){
+    return (
+        <div className='todo-list'>
+            {props.todos.map(todo =>
+                <Todo
+                    key={todo.task}
+                    todo={todo}
+                    />
+                )}
```
johnoro: The indentation here if off a bit; don't forget to use [prettier](https://prettier.io) for prettier code!

src/App.js
```diff
@@ -1,20 +1,28 @@
...
+    id: 2,
     completed: false
   }
 ];
 
+let nextId = 2;
+
+const getNewId = () => {
+  nextId += 1;
+  return nextId;
```
johnoro: You could shorten this with the pre-increment operator (which will increment a variable and evaluate to the new value, just like `numVar += 1`):
```javascript
const getNewId = () => ++nextId;
```

src/App.js
```diff
@@ -42,21 +50,50 @@ constructor(){
...
    });
  }
+
+  markComplete = id =>
+    this.setState({
+      todos: this.state.todos.map(todo =>{
+        if(todo.id === id){
+          return{
+            ...todo,
+            completed: todo.completed === false ? true : false
```
johnoro: You could just use the not/bang operator here:
```javascript
completed: !todo.completed
```

PR #678, opened 2018-11-09
David freitag

freitagdavid:


src/App.js
```diff
@@ -1,16 +1,85 @@
...
+        });
+    };
+
+    completeTask = id => {
+        this.setState({
+            tasks: this.state.tasks.map(task => {
+                if (task.id === id) {
+                    return {
+                        ...task,
+                        completed: task.completed === false ? true : false
```
johnoro: You could use the not/bang operator here:
```javascript
completed: !task.completed
```

src/components/TodoComponents/TodoForm.js
```diff
@@ -0,0 +1,23 @@
...
+                    placeholder="Add new task"
+                    value={props.inputText}
+                    onChange={props.handleChange}
+                />
+                <button type="submit">Add Todo</button>
+                <button type="button" onClick={props.clearTasks}>
+                    Clear Completed
+                </button>
+            </form>
+        </>
```
johnoro: Why use a fragment here when `form` is your parent already? It has no effect and is unnecessary!

PR #680, opened 2018-11-13
vanilla sub

alexcartaz:


src/less/global.less
```diff
@@ -0,0 +1,13 @@
+* {
+  box-sizing: border-box;
+}
+
+html, body {
+  margin: 0; 
+  padding: 0;
+  font-size: 62.5%;
```
johnoro: Having this applied to the html and the body seems to make the font-size extra small; it could use some up-sizing so that I can see the list items better!

src/App.js
```diff
@@ -1,16 +1,84 @@
...
   render() {
+    //console.log(this.state.toDoList);
+    var self = this;
     return (
-      <div>
-        <h2>Welcome to your Todo App!</h2>
+      <div className="App">
+        <h2>To Do List: MVP</h2>
+        {this.state.toDoList.map(item => {
+          return(<div key={item.id} onClick={() => self.changeStatus(item.id)} className={item.status && "strikethrough"}>{item.task}</div>);
```
johnoro: I'd recommend breaking this up for readability:
```javascript
{this.state.toDoList.map(item => {
  return (
    <div 
      key={item.id} 
      onClick={() => self.changeStatus(item.id)} 
      className={item.status && "strikethrough"}
    >
      {item.task}
    </div>
  );
})}
```

src/App.js
```diff
@@ -1,16 +1,84 @@
...
+  }
+
+
+  
   // you will need a place to store your state in this component.
   // design `App` to be the parent component of your application.
   // this component is going to take care of state, and any change handlers you need to work with your state
   render() {
+    //console.log(this.state.toDoList);
+    var self = this;
```
johnoro: Why do this reassignment? It referring to the same thing AND is the same length, so there's not a lot of point in it that I'm able to see.

src/App.js
```diff
@@ -1,16 +1,84 @@
...
     return (
-      <div>
-        <h2>Welcome to your Todo App!</h2>
+      <div className="App">
+        <h2>To Do List: MVP</h2>
+        {this.state.toDoList.map(item => {
+          return(<div key={item.id} onClick={() => self.changeStatus(item.id)} className={item.status && "strikethrough"}>{item.task}</div>);
+        })}
+        <input type="text" name="toDo" placeholder="...to do" value={this.state.inputText} onChange={this.updateCurrentTodo}></input>
```
johnoro: Break this up too!

PR #704, opened 2018-12-05
To Do List Kseniya Platonava

Kseniyapl:
@TheBrockStar

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,16 +1,60 @@
-import React from 'react';
+import React, { Component } from 'react'
+//import './App.css'
+import TodoList from './components/TodoComponents/TodoList'
+import TodoForm from './components/TodoComponents/TodoForm'
 
-class App extends React.Component {
-  // you will need a place to store your state in this component.
-  // design `App` to be the parent component of your application.
-  // this component is going to take care of state, and any change handlers you need to work with your state
+class App extends Component {
+  inputElement = React.createRef()
+  constructor() {
+    super()
+    this.state = {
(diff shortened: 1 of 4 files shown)


src/components/TodoComponents/TodoList.js
```diff
@@ -1,2 +1,28 @@
 // your components will all go in this `component` directory.
 // feel free to change this component.js into TodoList.js
+import React, { Component } from 'react'
+
+class TodoList extends Component {
+  componentDidUpdate() {
+    this.props.inputElement.current.focus()
+  }
```
TheBrockstar: What problem were you running into that led you to implement this?

src/App.js
```diff
@@ -1,16 +1,60 @@
...
+      items: [],
+      currentItem: {
+        text: '',
+        key: '',
+      },
+    }
+  }
+  deleteItem = key => {
+    const filteredItems = this.state.items.filter(item => {
+      return item.key !== key
```
TheBrockstar: This is a good way of handling deleting items from state. However, the MVP was for you to be able to set the state of each todo so that a "completed" property is set to true or false when each item is clicked. Then you are supposed to be able to clear all of the "completed" todo items. Currently, this functionality isn't present in your app.

Please take a look at this repo: https://codesandbox.io/s/github/lilmatchamonster/React-Todo

and try to refactor your todo list so that it handles completing and deleting todos in a similar way. The example above is one of the only projects that I've seen that got the functionality completely right.

PR #706, opened 2018-12-05
react-js

goonerfari:
Todo List app

@thebrockstar


src/App.js
```diff
@@ -1,13 +1,101 @@
...
+    else {
+      currentTodos.push(newItem);
+    }
+    this.setState({todos: currentTodos});
+
+  }
+
+  markComplete = (todo) => {
+    
+    const htmlElementsArray = document.querySelectorAll('li');
```
TheBrockstar: Directly interacting with the DOM in this way is a React Antipattern.

Instead, make sure that you are taking advantage of your event listeners by passing you event through, so that you can interact with its specific target.

ie.
**On your Todo component:**
`onClick={e => props.checkTodo(e, props.todo)}`

**On App.js**:
`markComplete = (e, todo) => {`
`currentTodo = e.currentTarget`

An even BETTER way is to have the class entirely controlled by state. That is a TRUE React way of handling something like this. Here is a small example: 

`className={props.todo.completed ? "completed" : ""}`

So you update the class by updating the completed property of the respective item in state.

src/App.js
```diff
@@ -1,13 +1,101 @@
...
+        }
+      ],
+      todo: "",
+      completedList: []
+    } 
+  }
+
+  addTodo = (text) => {
+
+    const currentTodos = this.state.todos.slice();
```
TheBrockstar: Good! This method is being handled well. Very REACTive ;)

src/components/TodoComponents/TodoList.js
```diff
@@ -1,2 +1,24 @@
...
 // feel free to change this component.js into TodoList.js
+
+import React from 'react';
+import Todo from './Todo';
+
+const TodoList = props => {
+    
+    return (
+        <ul>
+            {props.todos.map(todo => {
```
TheBrockstar: Good! Mapping through a property passed down from a parent component's state is the best way to handle this.

src/App.js
```diff
@@ -1,13 +1,101 @@
...
+      if (htmlElementID === todoID) {
+        htmlElementsArray[i].classList.toggle('completed');
+      }
+    }
+  }
+
+  clear = () => {
+    const cleared = this.state.todos.slice();
+    const newArray = [];
+    cleared.filter(clearItem => {
```
TheBrockstar: Excellent job here. `.filter()` is the best method for bulk removing items from state in this way. Well done!

PR #719, opened 2018-12-07
Test ToDo Day2

Kseniyapl:
@TheBrockStar

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,16 +1,60 @@
-import React from 'react';
+import React, { Component } from 'react'
+//import './App.css'
+import TodoList from './components/TodoComponents/TodoList'
+import TodoForm from './components/TodoComponents/TodoForm'
 
-class App extends React.Component {
-  // you will need a place to store your state in this component.
-  // design `App` to be the parent component of your application.
-  // this component is going to take care of state, and any change handlers you need to work with your state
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -1,16 +1,72 @@
...
+      currentItem: {
+        text: '',
+        key: '',
+        completed:false
+      },
+    }
+  }
+  deleteItem = key => {
+    var filteredItems = this.state.items.filter(item => {
+      return item.key === key
```
TheBrockstar: I wouldn't have thought to use `.filter()` here, but it totally works. How would you do it with `.map()`

What if someone completed a task by accident? How could you make it so that, after a click, completed would be set to false if the Todo had already been completed?

src/App.js
```diff
@@ -1,16 +1,72 @@
...
+
+    })
+    filteredItems [0].completed = true
+    this.setState({
+      items: this.state.items,
+    })
+  }
+  deleteAll = e => {
+    var filteredItems = this.state.items.filter(item => {
+      return !item.completed
```
TheBrockstar: This is great! Well done using the `.filter()` method. That is the best way to delete multiple items out of state (assuming their is something common among them that you can filter by).

